PR #27, opened 2022-01-21
Avoid calling asyncio run from within a running event loop

### LINKED ISSUE #26: Make datacache safe to call from within a running asyncio event loop

jesper-friis:
Closes #26 

oteapi-services runs from an asyncio event loop, but asyncio.run() cannot be called from a running event loop.

This PR checks if we have a running event loop. If so, the datacache methods will create a asyncio coroutine and submit it to the event loop. Thanks Casper for the solution.

This is a blocking for oteapi-services

### DIFF
--- a/oteapi/datacache/datacache.py
+++ b/oteapi/datacache/datacache.py
@@ -77,7 +77,15 @@ def asyncrun(func, *args) -> "Any":
         result = await future
         return result
 
-    return asyncio.run(async_func(args))
+    try:
+        # Will raise a RuntimeError if there is no event loop
+        asyncio.get_running_loop()
+    except RuntimeError:
+        # no event loop
+        return asyncio.run(async_func(args))
+    else:
+        # within a running event loop
+        return await func(*args)
 
 
 class DataCache:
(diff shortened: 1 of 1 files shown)


oteapi/datacache/datacache.py
```diff
@@ -77,7 +77,15 @@ async def async_func(args):
...
-    return asyncio.run(async_func(args))
+    try:
+        # Will raise a RuntimeError if there is no event loop
+        asyncio.get_running_loop()
+    except RuntimeError:
+        # no event loop
+        return asyncio.run(async_func(args))
+    else:
+        # within a running event loop
+        return func(*args)
```
CasperWA, line 88: If a loop is already running, couldn't we simply run the function in `loop.run_in_executor(None, func, *args)`? Making it run in the current loop?

CasperWA (changes requested):
Offline we discussed to completely remove the `asyncrun` functionality and instead return the function calls as is.

jesper-friis:
Replaced by PR #32

PR #29, opened 2022-01-22
Setup docs framework and implement creating any strategy function

### LINKED ISSUE #9: Setup documentation framework

### LINKED ISSUE #13: Attempt simplifying strategy factory function
The `create_X_strategy()` functions in each interface module could, in theory, be called through a single function, which can then be the only public functionality necessary for creating strategies.

CasperWA:
Closes #9 

Set up MkDocs documentation framework.
Have special top names to collect all pydantic models and core strategies.
Add doc-strings for all classes, functions, and methods where applicable.
Rename `DownloadConfig` to `DataCacheConfig`, since the download strategies use the `ResourceConfig`.

Closes #13 

This PR also implements a wrapper function for creating *any* strategy.


CasperWA:
The documentation has been setup already to the `gh-pages` branch. It will be auto-updated upon every new push to `master` as well as for every new release.

README.md
```diff
@@ -2,22 +2,6 @@
 
 > Framework for accessing data resources, mapping data models, describing the data to ontologies and perform data transformations
 
-## Content
```
jesper-friis, line 5: What is the reason to remove the content? I think it is useful for people viewing the project on GitHub - which is the entrypoint for most people - even though it is a small burden to keep updated...

CasperWA: Since there aren't a lot of things on the main page/README and I'd rather have people use the documentation, I chose to remove it as it is redundant in the documentation. The Table of Contents is already available on the right side in the documentation theme. If you feel it's absolutely necessary I can return it, but then I'd feel forced to try and dynamically remove it when creating the `index.md` for the documentation.
On the other hand, we can add a sentence stating the "we highly recommend people to read this README through [the documentation](https://EMMC-ASBL.github.com/oteapi-core)"?

oteapi/strategies/download/https.py
```diff
@@ -9,25 +8,30 @@
...
-    def initialize(
-        self, session: "Optional[Dict[str, Any]]" = None
-    ) -> "Dict[str, Any]":
-        """Initialize"""
+    def initialize(self, **_) -> "Dict[str, Any]":
+        """Initialize."""
         return {}
 
-    def get(self, session: "Optional[Dict[str, Any]]" = None) -> "Dict[str, Any]":
+    def get(self, **_) -> "Dict[str, Any]":
```
jesper-friis, line 34: I am not fully comfortable with this use of anonymous arguments. The get() and other methods should expect one optional `session` parameter, not an arbitrary number of anonymous keyword arguments.

oteapi/strategies/filter/crop_filter.py
```diff
@@ -1,35 +1,41 @@
...
-    def initialize(
-        self, session: "Optional[Dict[str, Any]]" = None
-    ) -> "Dict[str, Any]":
-        """Initialize strategy and return a dictionary"""
+    def initialize(self, **_) -> "Dict[str, Any]":
+        """Initialize strategy and return a dictionary."""
         return {"result": "collectionid"}
 
-    def get(self, session: "Optional[Dict[str, Any]]" = None) -> "Dict[str, Any]":
+    def get(self, **_) -> "Dict[str, Any]":
```
jesper-friis, line 38: General comment: I am not fully comfortable with this use of anonymous keyword arguments. The `get()` and several other methods takes one argument `session` as input. But it is not all implementations that uses it. I think it is better to be explicit about that.

jesper-friis (approved):

oteapi/strategies/download/https.py
CasperWA: It's fair enough. Discussing this offline, we have arrived at the conclusion of keeping in `session` explicitly - even if not used - to ensure the signature is clear for external strategy developers as well.
The pylint disable statements must be returned also.

README.md
CasperWA: I have added this suggested sentence to the README.

jesper-friis (approved):

PR #47, opened 2022-01-26
Start implementing entry point logic

### LINKED ISSUE #10: Extend and "safeguard" plugin loading through entry points
There are currently some issues concerning registering strategies for the same schemes - both from different "unique" sources, but also within the same module being loaded.
There may be other edge cases to be considered and "safeguarded" against.

### LINKED ISSUE #21: Lazy strategy loading
The following is a lazy loading implementation suggestion (it came up in a discussion between me and @jesper-friis).

If the setuptools' entry_points values focus on the strategy schemes instead of the strategy types, it may allow lazy loading of the modules hosting the strategies.

Instead of the current syntax of the entry_points (as defined in a `setup.cfg` file):

```ini
[options.entry_points]
oteapi.download_strategy =
  core.file = oteapi.strategies.download.file
  core.https = oteapi.strategies.download.https
  core.sftp = oteapi.strategies.download.sftp
oteapi.filter_strategy =
  core.crop = oteapi.strategies.filter.crop_filter
  # ...
```

It can be changed to something like:

```ini
[options.entry_points]
oteapi.scheme =
  http = oteapi.strategies.download.https
  https = oteapi.strategies.download.https
oteapi.mediaType =
  image.jpg = oteapi.strategies.parse.image_jpeg
  image.jpeg = oteapi.strategies.parse.image_jpeg
  image.j2p = oteapi.strategies.parse.image_jpeg
  # ...
```

This will make it possible for the plugins to declare what scheme/mediaType+value they are implementing strategies for (as well as where to find the strategy implementations).

Now, when loading the OTE-API application, the entry points group (e.g., `oteapi.scheme`) + name (e.g., `https`) uniquely defines a valid value for `StrategyFactory.register()`.
Hence, with the proper Python API for loading the entry points, the actual module import (in the given example of `oteapi.strategies.download.https`) can be postponed until the (`scheme`, `https`)-strategy is requested.

Since `importlib.import_module` refers back to built-in functions that always check whether the module exists in `sys.modules` before loading it, calling `importlib.import_module` for the same module several times is not expensive, and will merely return the cached module.

Finally, to ease development for plugin-developers, this list of entry points can be generated from functionality existing in `oteapi-core` (not yet written). The functionality could be implemented as part of the plugin repository's `setup.py` or as a `pre-commit` hook - or both.
In this way, it should be easier for plugin-developers to get started.
It's understood that the functionality delivered from the side of `oteapi-core` should include ways of excluding certain strategy implementations to be findable, etc. Essentially achieving the same amount of control currently available through the manual addition of `entry_points` entries.

### LINKED ISSUE #57: Consider removing the `create_*_strategy()` functions
We can use the `create_strategy()` function instead.

CasperWA:
Closes #21 
Closes #10 

This PR will extend the plugins module with stricter and more "secure" handling of loading plugins through entry points.
It will also update the entry point syntax in order to facilitate lazy loading of the entry points - only ever importing a strategy when it is needed and called, avoiding having to auto-load all strategies when writing a server service for the OTE-API.

Still to do:

- [X] Finalize the new logic/code.
- [X] Make current tests pass.
- [x] Write new tests for the new code.

New closures and fixes. See comments below for details.
Closes #57


CasperWA:
A **major shift** in the plugin system has been implemented.

By writing out the class name in the entry points the need to explicitly register strategies has been removed.
Instead one should duplicate the module+class name in an entry point to define all the strategies a single class implements.

The classes are then *only* imported when the `StrategyFactory.make_strategy()` is invoked. Which is usually done through the `create_*_strategy()` functions.

The new implementation also provides much better granular control of the entry points representing strategies.

CasperWA:
Tests for the new code is still missing.

oteapi/plugins/factories.py
```diff
@@ -104,70 +111,81 @@ class StrategyFactory:
...
+        Returns:
+            The strategy name provided in the configuration.
 
-    @classmethod
-    def unregister(cls, *args: "Tuple[str, ValueType]") -> None:
-        """Unregister a strategy."""
-        for index in args:
-            cls.strategy_create_func.pop(index, None)
+        """
+        if strategy_type == StrategyType.DOWNLOAD:
```
lovfall, line 180: This looks like a hack to me. Why do you have to do a special trick for the download strategy here?

CasperWA: This was also done previously.
Unfortunately, where the strategy type + name is derived directly from a key + value in the configuration normally, this is not the case for the download strategy, where the strategy type (`scheme`) is a special attribute of the `downloadUrl` configuration key. The strategy name is the value of this `scheme` attribute.

CasperWA: I have created this special extra method to deal with these edge cases - instead of making it a proper "hack" by simply implementing it directly in the `make_strategy()` method.

Therefore, should any other strategy-specific quirks appear in the future, they can be added to this special (private) method.

CasperWA:
Added "BLOCKING" label because this PR blocks other PRs (especially #49).

oteapi/plugins/entry_points.py
```diff
@@ -0,0 +1,505 @@
...
+        return self.get_entry_point(key)
+
+    def get_entry_point(
+        self,
+        key: "Union[EntryPointStrategy, str, Tuple[Union[StrategyType, str], str]]",
+    ) -> EntryPointStrategy:
+        """Retrieve an entry point from the collection.
+
+        Raises:
+            KeyError: If an entry point cannot be found in the collection.
```
jesper-friis, line 394: Can also raise TypeError

oteapi/plugins/entry_points.py
```diff
@@ -0,0 +1,505 @@
...
+        return f"{self.__class__.__name__}(*{tuple(sorted(self._entry_points))!r})"
+
+
+def get_strategy_entry_points(
+    strategy_type: "Union[StrategyType, str]",
+    enforce_uniqueness: bool = True,
+) -> EntryPointStrategyCollection:
+    """Retrieve all entry points from a specific strategy type.
+
+    Raises:
```
jesper-friis, line 471: Is there a conventional order for the fields in google docstrings? For me it feels more natural that Raises comes after Return. 

This comment applies for other docstrings as well.

oteapi/plugins/entry_points.py
```diff
@@ -0,0 +1,505 @@
...
+    @lru_cache
+    def map_to_field(self) -> str:
+        """Map enumeration value to the strategy type's field."""
+        return {
+            "download": "scheme",
+            "filter": "filterType",
+            "mapping": "mappingType",
+            "parse": "mediaType",
+            "resource": "accessService",
+            "transformation": "transformation_type",
```
jesper-friis, line 62: For consistency in with dcat and other filter type names, I think we should wrhte "transformationType" here (and everywhere else in the code).

jesper-friis (commented):
Great PR. Added some comments.

oteapi/plugins/entry_points.py
```diff
@@ -0,0 +1,505 @@
...
+    """
+
+    DOWNLOAD = "download"
+    FILTER = "filter"
+    MAPPING = "mapping"
+    PARSE = "parse"
+    RESOURCE = "resource"
+    TRANSFORMATION = "transformation"
+
+    @lru_cache
```
jesper-friis, line 53: What is the purpose of this cache? The dict lookup below is really fast and shouldn't need any cache...

oteapi/strategies/download/file.py
```diff
@@ -34,7 +33,6 @@ class FileConfig(BaseModel):
 
 
 @dataclass
-@StrategyFactory.register(("scheme", "file"))
```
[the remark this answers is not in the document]
CasperWA: Yes.

setup.cfg
```diff
@@ -1,16 +1,24 @@
...
-  core.sql_query = oteapi.strategies.filter.sql_query_filter
-oteapi.parse_strategy =
-  core.excel.xlsx = oteapi.strategies.parse.excel_xlsx
-  core.image.jpeg = oteapi.strategies.parse.image_jpeg
-  core.sqlite = oteapi.strategies.parse.application_vnd_sqlite
-  core.text.csv = oteapi.strategies.parse.text_csv
-  core.text.json = oteapi.strategies.parse.text_json
-oteapi.transformation_strategy =
-  core.celery = oteapi.strategies.transformation.celery_remote
+oteapi.scheme =
```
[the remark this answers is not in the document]
CasperWA: We do not want to do this as previously discussed. The upload should be a download strategy - we should just rename them. But they should definitely be `download` etc. But I think they already are? This seems to be an outdated commit you're commenting on?

oteapi/plugins/entry_points.py
```diff
@@ -0,0 +1,505 @@
...
+            "download": "scheme",
+            "filter": "filterType",
+            "mapping": "mappingType",
+            "parse": "mediaType",
+            "resource": "accessService",
+            "transformation": "transformation_type",
+        }[self.value]
+
+    @classmethod
+    @lru_cache
```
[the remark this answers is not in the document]
CasperWA: True - I'm trying to start learning to use it and thought it was nice? But perhaps not?

oteapi/plugins/entry_points.py
CasperWA: I think my original idea here is to have it store the dictionary in the cache instead of creating it anew each time the method is called? But maybe that's irrelevant?

oteapi/plugins/entry_points.py
CasperWA: dcat does not have a `transformationType`. Nor does it have `mappingType` or `filterType`. But I guess we should be consistent, yes. Hmm.... fiiiine :)

oteapi/plugins/entry_points.py
```diff
@@ -0,0 +1,505 @@
...
+        This method allows one to also initialize a StrategyType with an actual
+        strategy type string, e.g., `scheme`, `mediaType`, etc.
+
+        Raises:
+            ValueError: As normal if the enumeration value is not valid.
+
+        """
+        try:
+            return cls.map_from_field(value)
+        except KeyError:
```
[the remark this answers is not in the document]
CasperWA: No, no. This is if the method `map_from_field()` should raise a `KeyError`, then I catch it and instead try to initialize a `StrategyType` Enum by invoking the class itself, which may then raise `ValueError` as documented in the docstring.

oteapi/plugins/entry_points.py
```diff
@@ -0,0 +1,505 @@
...
+            return cls(value)
+
+    @classmethod
+    @lru_cache
+    def all_values(cls) -> "Tuple[str, ...]":
+        """Return all values."""
+        return tuple(_.value for _ in cls)
```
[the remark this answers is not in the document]
CasperWA: No need to use a proper name here and the variable is not used for anything other than an iteration variable. This is standard Python practice - using `_` as an anonymous variable. At least to me, but perhaps not?

oteapi/plugins/entry_points.py
```diff
@@ -0,0 +1,505 @@
...
+
+        match = self.ENTRY_POINT_NAME_REGEX.match(self._entry_point.name)
+        if match is None:
+            raise ValueError(
+                "Could not determine package name and/or strategy name for entry "
+                f"point: {self.full_name}"
+            )
+
+        self._match = match
+        self._type = StrategyType(self._entry_point.group[len("oteapi.") :])
```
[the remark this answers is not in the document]
CasperWA: No. This looks at the entry point _group_, i.e., either `oteapi.download`, `oteapi.parse`, ... Not the actual entry points listed under a group.

oteapi/plugins/entry_points.py
CasperWA: Okay - for me it was more natural the the last section was `Returns`, and when getting there the reader is already aware what exceptions may occur - but I can switch it up for sure if you want. I think the rendering in the documentation doesn't care either way.

oteapi/plugins/entry_points.py
CasperWA: And if I understand `lru_cache` correctly, then it merely caches the return value based on the input parameters - so eventually it will have cached the whole dictionary.

setup.cfg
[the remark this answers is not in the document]
jesper-friis: That was an old comment, not sure why it shows 15h ago. Resolving this conversation

oteapi/plugins/entry_points.py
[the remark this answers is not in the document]
jesper-friis: I had to think a lot before I understood what the anonymous `_` stands for. A proper name for the variable makes the code much easier to read.

oteapi/plugins/entry_points.py
jesper-friis: It is probably because I have earlier been using [numpydoc](https://numpydoc.readthedocs.io/en/latest/format.html), which defines a very specific order of the different section. But I don't think it hurts to use the order they have defined

oteapi/plugins/entry_points.py
jesper-friis: In this case it is not a big deal. But the `lru_cache` is essentially a dict with some additional logics for eviction++. Since what it is caching is already a dict, it just adds an unnessary additional layer without any benefits.

oteapi/plugins/factories.py
jesper-friis: I think this is a good way to do it.

oteapi/plugins/entry_points.py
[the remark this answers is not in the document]
jesper-friis: Caching make sense when the underlying calculation is expensive. A dict lookup is cheap (actually cheaper than the cache since in include additional logics). Hence, a cache make no sense in this case.

oteapi/plugins/entry_points.py
jesper-friis: It could be a low priority new issue to not slow down this wonderful PR any further...

oteapi/plugins/entry_points.py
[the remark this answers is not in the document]
CasperWA: okay

oteapi/plugins/entry_points.py
CasperWA: Good suggestion, as this requires a bit of manual work to go through all the code and cannot be done with a simple search-and-replace. Please make another issue for this.

jesper-friis (approved):

oteapi/plugins/entry_points.py
CasperWA: Opened #58

Issue #34, opened 2022-01-24
Configuration bug


TorgeirUstad:
After commit dc1ec0d, creating an ImageDataParseStrategy fails with message

`TypeError: 'DataCacheConfig' object is not subscriptable`

on line 45 in "image_jpeg.py".

CasperWA:
The issue is the mixed type for `configuration` - specifically for the `ResourceConfig` model: `Union[DataCacheConfig, Dict[str, Any]`.

There are two solutions in my opinion:
1. Create a utility function to "get" attributes from an object that may be either a pydantic model or a dictionary.
2. Use one or the other throughout for the repeating `configuration` attribute in the configs. Preferrably use pydantic models - one can simulate the flexibility of a dictionary by allowing the addition of extra attributes.

I personally prefer the latter solution.

PR #53, opened 2022-01-31
Add some dictionary functionality to configuration models

TorgeirUstad:
This is an implementation of the second approach suggested by Casper in his comment on issue #34.

### DIFF
--- a/oteapi/datacache/datacache.py
+++ b/oteapi/datacache/datacache.py
@@ -109,10 +109,11 @@ class DataCache:
             cache_dir = self.config.cacheDir
         if isinstance(cache_dir, str):
             cache_dir = Path(cache_dir)
-        if cache_dir.is_absolute():
+        if cache_dir.is_absolute():  # type: ignore[union-attr]
             self.cache_dir = cache_dir
         else:
-            self.cache_dir = Path(tempfile.gettempdir()).resolve() / cache_dir
+            temp_path = Path(tempfile.gettempdir()).resolve()
+            self.cache_dir = temp_path / cache_dir  # type: ignore[operator]
 
         self.diskcache = DiskCache(directory=self.cache_dir)
 
@@ -169,7 +170,7 @@ class DataCache:
             expire = self.config.expireTime
 
         self.diskcache.set(key, value, expire=expire, tag=tag)
-        return key
+        return key  # type: ignore[return-value]
 
     def get(self, key: str) -> "Any":
         """Return the value corresponding to `key`.
--- a/oteapi/models/datacacheconfig.py
+++ b/oteapi/models/datacacheconfig.py
@@ -1,10 +1,12 @@
 """Pydantic DataCache Configuration Data Model."""
 from pathlib import Path
 
-from pydantic import BaseModel, Field
+from pydantic import Field
+
+from oteapi.models.genericconfig import GenericConfig
 
 
-class DataCacheConfig(BaseModel):
+class DataCacheConfig(GenericConfig):
     """DataCache Configuration."""
 
     cacheDir: Path = Field("oteapi", description="Cache directory.")
--- a/oteapi/models/genericconfig.py
+++ b/oteapi/models/genericconfig.py
@@ -1,0 +1,60 @@
+"""Generic data model for configuration attributes."""
+from typing import Any
+
+from pydantic import BaseModel
+
+
+class GenericConfig(BaseModel):
+    """Generic class for configuration objects.
+
+    The goal of this class is that objects of the class can be treated
+    both like a BaseModel and like a dictionary, e.g. subscripted.
+    """
+
+    def __init__(self, **kwargs) -> None:
+        """Initializer from dictionary."""
+        super().__init__(**kwargs)
+        for key, value in kwargs.items():
+            # Attributes have been set by the call to super() above
+            if not self.__contains__(key):
+                self.__dict__[key] = value
+
+    def __call__(self, *args, **kwargs) -> Any:
+        """Enable calling the function in args[0]."""
+        func = args[0]
+        if func in dir(self):
+            # This object has this function
+            found = getattr(self, func)
+        elif func in dir(dict):
+            # The dictionary representation has the function
+            found = getattr(dict(self), func)
+        else:
+            raise AttributeError(func)
+        return found(*args[1:], **kwargs)
+
+    def __contains__(self, name: Any) -> bool:
+        """Enable using the 'in' operator on this object."""
+        return self.__dict__.__contains__(name)
+
+    def __delitem__(self, key: str) -> None:
+        """Enable deletion access through subscription."""
+        if key in dir(self):
+            self.__delattr__(key)
+        else:
+            raise KeyError(key)
+
+    def __getitem__(self, key: str) -> Any:
+        """Enable read access through subscription."""
+        if key in dir(self):
+            return getattr(self, key)
+        raise KeyError(key)
+
+    def __setitem__(self, key: str, value: Any) -> None:
+        """Enable write access through subscription."""
+        if key in self.__dict__:
+            target_type = type(self.__dict__[key])
+            if not value.isinstance(target_type):
+                raise TypeError(
+                    "Mapped value must be subclass of " + target_type.__name__
+                )
+        self.__dict__[key] = value
--- a/oteapi/models/resourceconfig.py
+++ b/oteapi/models/resourceconfig.py
@@ -1,12 +1,13 @@
 """Pydantic Resource Configuration Data Model."""
-from typing import Any, Dict, Optional, Union
+from typing import Any, Dict, Optional
 
-from pydantic import AnyUrl, BaseModel, Field, root_validator
+from pydantic import AnyUrl, Field, root_validator
 
 from oteapi.models.datacacheconfig import DataCacheConfig
+from oteapi.models.genericconfig import GenericConfig
 
 
-class ResourceConfig(BaseModel):
+class ResourceConfig(GenericConfig):
     """Resource Strategy Data Configuration.
 
     Important:
@@ -70,7 +71,7 @@ class ResourceConfig(BaseModel):
         None,
         description="The entity responsible for making the resource/item available.",
     )
-    configuration: Union[DataCacheConfig, Dict[str, Any]] = Field(
+    configuration: DataCacheConfig = Field(
         {},
         description="Resource-specific configuration options given as key/value-pairs.",
     )
--- a/docs/api_reference/models/genericconfig.md
+++ b/docs/api_reference/models/genericconfig.md
@@ -1,0 +1,5 @@
+# genericconfig
+
+::: oteapi.models.genericconfig
+    rendering:
+      show_if_no_docstring: true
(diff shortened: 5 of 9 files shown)


oteapi/datacache/datacache.py
```diff
@@ -109,10 +109,11 @@ def __init__(
             cache_dir = self.config.cacheDir
         if isinstance(cache_dir, str):
             cache_dir = Path(cache_dir)
-        if cache_dir.is_absolute():
+        if cache_dir.is_absolute():  # type: ignore[union-attr]
             self.cache_dir = cache_dir
         else:
-            self.cache_dir = Path(tempfile.gettempdir()).resolve() / cache_dir
+            temp_path = Path(tempfile.gettempdir()).resolve()
+            self.cache_dir = temp_path / cache_dir  # type: ignore[operator]
```
CasperWA, line 116: Is there a reason to split these lines up in two? If not, please return them to a single line.

oteapi/datacache/datacache.py
```diff
@@ -169,7 +170,7 @@ def add(
             expire = self.config.expireTime
 
         self.diskcache.set(key, value, expire=expire, tag=tag)
-        return key
+        return key  # type: ignore[return-value]
```
CasperWA, line 173: This suggests `self.config.accessKey` or the `gethash()` function may not return a `str` always...

CasperWA (changes requested):
Looks good - but I see that you misunderstood me.
My intent was to make the `configuration` attribute/field of the `GenericConfig` type - maybe calling it `AttrDict` instead (short for attribute dictionary)?

oteapi/datacache/datacache.py
TorgeirUstad: The problem was that mypy claimed that `cache_dir` was `None`, and thus that the `/` operator was invalid. When I added the ignore statement to the end of the line, it became too long (a different complaint), so I split it up.

TorgeirUstad:
I just added a bugfix to the existing `GenericConfig` code, in case we want to keep it.

oteapi/datacache/datacache.py
```diff
@@ -109,10 +109,11 @@ def __init__(
             cache_dir = self.config.cacheDir
         if isinstance(cache_dir, str):
             cache_dir = Path(cache_dir)
-        if cache_dir.is_absolute():
+        if cache_dir.is_absolute():  # type: ignore[union-attr]
```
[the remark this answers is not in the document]
CasperWA: I think for that specific use case I would make `DataCacheConfig` a subclass of `GenericConfig` - but yes - otherwise it seems you've caught my meaning.
And you should definitely keep the existing `GenericConfig`, but move all the methods you've made to a new class called `AttrDict` or similar. Also - there is actually no need to create a fancy class, since a simple pydantic model will work the same - in a sense. If you simply use the internal class `Config` to set `extra = "allow"` (see the [docs here](https://pydantic-docs.helpmanual.io/usage/model_config/)). Then one can instantiate the `configuration` parameter using a dictionary and always retrieve the keys either as an attribute or via the built-in `getattr()` function, where a default value can be included.

oteapi/datacache/datacache.py
CasperWA: Yes, it's because mypy thinks `cache_dir` is still either `None` or `Path`, and `None` doesn't have any of these functionalities. Hmm - it's weird. But fair.

oteapi/datacache/datacache.py
CasperWA: I think that's because of the `self.config.accessKey` type. This means we should make sure a `str` is returned and `key` is not `None` when returned from this method.

oteapi/models/genericconfig.py
```diff
@@ -0,0 +1,60 @@
...
+        extra = "allow"
+
+    def __contains__(self, name: Any) -> bool:
+        """Enable using the 'in' operator on this object."""
+        return self.__dict__.__contains__(name)
+
+    def __delitem__(self, key: str) -> None:
+        """Enable deletion access through subscription."""
+        if key in dir(self):
+            self.__delattr__(key)
```
CasperWA, line 23: This is tricky. Since it's a pydantic model, there are some other internal `BaseModel` attributes that should be updated alongside this. I'm unsure whether calling `__delattr__` will update these.
The ones I'm thinking of are `__fields__` and `__fields_set__` (or something to that effect).

This might be fine, but having a test for these methods is crucial, to ensure they properly update the object as is intended.

oteapi/models/genericconfig.py
```diff
@@ -0,0 +1,60 @@
...
+        raise KeyError(key)
+
+    def __setitem__(self, key: str, value: Any) -> None:
+        """Enable write access through subscription."""
+        if key in self.__dict__:
+            target_type = type(self.__dict__[key])
+            if not isinstance(value, target_type):
+                raise TypeError(
+                    "Mapped value must be subclass of " + target_type.__name__
+                )
```
CasperWA, line 40: This is interesting. I really like this section. I'm still wondering about some mixes between this and pydantic, but tests should answer those questions. Great!

oteapi/models/genericconfig.py
```diff
@@ -0,0 +1,60 @@
...
+        self.__dict__[key] = value
+
+
+class GenericConfig(AttrDict):
```
CasperWA, line 44:
```suggestion
class GenericConfig(BaseModel):
```
Not sure I'm for making the `GenericConfig` a wannabe dictionary. However, I'm not completely against it either. But for sure it shouldn't allow random extras... I think. That's for the `configuration` only.
Therefore, I think it's easier to use `BaseModel` here; we want the strict attribute validation for the config classes.

oteapi/models/genericconfig.py
```diff
@@ -0,0 +1,60 @@
...
+    )
+
+    description: str = Field(
+        __doc__,
+        description="A description of the configuration model.",
+    )
+
+    @classmethod
+    def __init_subclass__(cls) -> None:
+        cls.__fields__["description"].default = cls.__doc__
```
CasperWA, line 60: Elegant

oteapi/models/datacacheconfig.py
```diff
@@ -1,10 +1,12 @@
...
 from pathlib import Path
 
-from pydantic import BaseModel, Field
+from pydantic import Field
 
+from oteapi.models.genericconfig import GenericConfig
 
-class DataCacheConfig(BaseModel):
+
+class DataCacheConfig(GenericConfig):
```
CasperWA, line 9:
```suggestion
class DataCacheConfig(AttrDict):
```
It's a bit misleading, but this configuration class is actually not for a strategy, but for a strategy configuration's `configuration` attribute......

oteapi/models/datacacheconfig.py
```diff
@@ -1,10 +1,12 @@
 """Pydantic DataCache Configuration Data Model."""
 from pathlib import Path
 
-from pydantic import BaseModel, Field
+from pydantic import Field
 
+from oteapi.models.genericconfig import GenericConfig
```
CasperWA, line 6:
```suggestion
from oteapi.models.genericconfig import AttrDict
```

oteapi/models/resourceconfig.py
```diff
@@ -63,15 +64,12 @@ class ResourceConfig(BaseModel):
...
-    description: Optional[str] = Field(
-        None, description="A free-text account of the distribution."
-    )
     publisher: Optional[str] = Field(
         None,
         description="The entity responsible for making the resource/item available.",
     )
-    configuration: Union[DataCacheConfig, Dict[str, Any]] = Field(
-        {},
+    configuration: DataCacheConfig = Field(
```
CasperWA, line 71: When `DataCacheConfig` subclasses `AttrDict`, this makes sense. Otherwise it should be a `Union` of the two.

CasperWA (changes requested):
This is looking great!

There are still some things about `GenericConfig` versus `AttrDict`, but I've mentioned those in the comments below.

The main thing I'm worried about is the magic Python methods and how they affect the pydantic methods. This should be tested. Some very simple small sanity tests should do.

oteapi/models/genericconfig.py
TorgeirUstad: Thanks for the hint. I've updated the relevant attributes now. Now `del AttrDict["name"]` removes `name` from `AttrDict.__fields__` and `AttrDict.__fields_set__` (while `del AttrDict.name` doesn't) I plan on adding unit test(s) for the `Config` classes as well, if time allows for it.

oteapi/models/genericconfig.py
TorgeirUstad: Interesting indeed. Pydantic claims that

> pydantic enforces type hints at runtime

but without my code above, I was able to change DataCacheConfig.hashType (a string) to an int when subscripting. With my code, I got the expected exception. Pydantic didn't react when I changed the value with `DataCacheConfig.hashType =` either. I see that `BaseModel.__setattr__` is decorated with `@no_type_check`, so I added a `__setattr__` with type-checking to `AttrDict`.

oteapi/models/genericconfig.py
TorgeirUstad: If ´DataCacheConfig´ is the only `Config` class that will ever be used as a `configuration` field, then the other `Config` classes don't need this subscripting flexibility, so then `BaseModel` suffices for them. I've reverted `GenericConfig's` parent to `BaseModel`.

oteapi/models/genericconfig.py
TorgeirUstad: Now even more elegant - with a docstring!

oteapi/models/datacacheconfig.py
TorgeirUstad: If it will never be used as anything else, then ok. I've made the change and updated the docstring to mention this.

oteapi/models/datacacheconfig.py
TorgeirUstad: Done.

oteapi/models/resourceconfig.py
TorgeirUstad: `DataCacheConfig` did indirectly subclass `AttrDict` through `GenericConfig`, but now I've made the inheritance direct.

TorgeirUstad:
I have merged master into the branch now and solved a small conflict.

I was confused for a while because some tests from master failed (both in this branch and in master) in my local repository, but it turned out that the problem was my Python version. I was running 3.9.7, which apparently isn't new enough for the new master. Now I'm running 3.10.2, and everything works.

Maybe we should update the requirement on Python version? I haven't tested if 3.9.8 works, but the workflows do ok with 3.9.9.

oteapi/models/genericconfig.py
CasperWA: Would be great with the tests.
Just testing that the expected behavior is kept is enough. I can't speak for what people would expect when using `del`, but feel free to define that and state it in the tests (and PR). Then it is easier to spot for the reviewer and comment on eventually.

CasperWA (approved):
Looks good to me, cheers @TorgeirUstad 👍

PR #67, opened 2022-02-09
Made ResourceConfig.configuration a dict

jesper-friis:
Also made downloadUrl a union of AnyUrl and FileUrl.  It seems like a flaw in pydantic that AnyUrl doesn't include file urls, but that is actually the case...

### DIFF
--- a/oteapi/models/resourceconfig.py
+++ b/oteapi/models/resourceconfig.py
@@ -1,10 +1,9 @@
 """Pydantic Resource Configuration Data Model."""
-from typing import Any, Dict, Optional
+from typing import Any, Dict, Optional, Union
 
-from pydantic import AnyUrl, Field, root_validator
+from pydantic import AnyUrl, Field, FileUrl, root_validator
 
-from oteapi.models.datacacheconfig import DataCacheConfig
-from oteapi.models.genericconfig import GenericConfig
+from oteapi.models.genericconfig import AttrDict, GenericConfig
 
 
 class ResourceConfig(GenericConfig):
@@ -16,7 +15,7 @@ class ResourceConfig(GenericConfig):
 
     """
 
-    downloadUrl: Optional[AnyUrl] = Field(
+    downloadUrl: Optional[Union[AnyUrl, FileUrl]] = Field(
         None,
         description=(
             "Definition: The URL of the downloadable file in a given format. E.g. CSV "
@@ -68,8 +67,8 @@ class ResourceConfig(GenericConfig):
         None,
         description="The entity responsible for making the resource/item available.",
     )
-    configuration: DataCacheConfig = Field(
-        DataCacheConfig(),
+    configuration: Optional[AttrDict] = Field(
+        {},
         description="Resource-specific configuration options given as key/value-pairs.",
     )
 


oteapi/models/genericconfig.py
```diff
@@ -46,6 +46,26 @@ def __setitem__(self, key: str, value: Any) -> None:
...
+        """Return a view of all keys."""
+        return self.__dict__.keys()
+
+    def values(self):
+        """Return a view of all values."""
+        return self.__dict__.values()
+
+    def has_key(self, key):
+        """Return whether key is in self."""
+        return self.__dict__.has_key(key)
```
CasperWA, line 67: Not needed - `__contains__` already exists.
```suggestion
```

CasperWA (changes requested):
`AttrDict` needs more methods according to the [Python documentation](https://docs.python.org/3/library/collections.abc.html#collections-abstract-base-classes).
It should implement the methods required by `MutableMapping`, i.e., also those by `Mapping`.
Missing methods:
- [ ] `__iter__`
- [ ] `get`
- [ ] `__eq__`
- [ ] `__ne__`
- [ ] `pop`
- [ ] `popitem`
- [ ] `clear`
- [ ] `update`
- [ ] `setdefault`

oteapi/strategies/parse/excel_xlsx.py
```diff
@@ -21,7 +21,7 @@
 class XLSXParseDataModel(BaseModel):
     """Data model for retrieving a rectangular section of an Excel sheet."""
 
-    worksheet: str = Field(..., description="Name of worksheet to load.")
+    worksheet: str = Field("Sheet1", description="Name of worksheet to load.")
```
CasperWA, line 24: To set a default value? Will this value not change depending on the origin language for the created excel sheet? Like `Ark1` in Danish? Maybe this should be added to the description? Or perhaps instead some form of post-processing could be done to auto-choose the first sheet found in the excel file?

tests/models/test_genericconfig.py
```diff
@@ -0,0 +1,13 @@
+"""Tests for `oteapi.models.genericceconfig`"""
```
CasperWA, line 1:
```suggestion
"""Tests for `oteapi.models.genericconfig`"""
```

tests/models/test_genericconfig.py
```diff
@@ -0,0 +1,13 @@
...
+import pytest
+
+
+def test_attrdict() -> None:
+    """Test the behaviour of AttrDict."""
+    from oteapi.models.genericconfig import AttrDict
+
+    config = AttrDict(a=1, b='foo', c='bar')
+    assert config.a == 1
```
CasperWA, line 10:
```suggestion
    assert config.a == config[a] == config.get(a) == data[a]
```

tests/models/test_genericconfig.py
```diff
@@ -0,0 +1,13 @@
...
+    assert config.b == 'foo'
```
CasperWA, line 11:
```suggestion
    assert config.b == config[b] == config.get(b) == data[b]
```

tests/models/test_genericconfig.py
```diff
@@ -0,0 +1,13 @@
+"""Tests for `oteapi.models.genericceconfig`"""
+import pytest
+
+
+def test_attrdict() -> None:
+    """Test the behaviour of AttrDict."""
+    from oteapi.models.genericconfig import AttrDict
+
```
CasperWA, line 8:
```suggestion

    data = {"a": 1, "b": "foo", "c": "bar"}
```

tests/models/test_genericconfig.py
```diff
@@ -0,0 +1,13 @@
...
+    config = AttrDict(a=1, b='foo', c='bar')
```
CasperWA, line 9:
```suggestion
    config = AttrDict(**data)
```

tests/models/test_genericconfig.py
```diff
@@ -0,0 +1,13 @@
...
+    assert config.a == 1
+    assert config.b == 'foo'
+
+    assert **config == {'a': 1, 'b': 'foo', 'c': 'bar'}
```
CasperWA, line 13:
```suggestion
    assert **config == data
```

CasperWA (changes requested):

CasperWA (approved):
Cheers @jesper-friis

Issue #64, opened 2022-02-08
Be clear about the use of ResourceConfig.configuration


jesper-friis:
The `ResourceConfig.configuration` may be used by several configurations. For instance if you are instantiating the `excel_xlsx` parse strategy via the `file` download strategy, the `ResourceConfig.configuration` is used to configure the `XLSXParseDataModel`, `FileConfig` and `DataCacheConfig` data models. 

We need to decide how we want to deal with this. I think we have the following options:
1.  All the configuration data models share the same dict. This require that we ignore validation (by passing `extra=Extra.ignore` as argument to the config constructors), which makes it impossible to e.g. validate for spelling errors in the configurations.
2. Replace `ResourceConfig.configuration` with strategy-specific configurations:
```python
    datacacheConfig: DataCacheConfig = Field(DataCacheConfig(), "Datacache-specific configuration as key-value pairs.")
    downloadConfig: Dict = Field({}, "Download-specific configuration as key-value pairs.")
    parseConfig: Dict = Field({}, "Parse-specific configuration as key-value pairs.")
    ...  # this list may be extended with functionConfig, etc...
```
3. Keep `ResourceConfig.configuration` as a dict that should be of the form:
```python
{
    "datacacheConfig": {...}, 
    "downloadConfig": {...},
    "parseConfig": {...}, 
}
```
such that the different strategies can check it and pick out the configuration that it matches.

Personally I would prefer option 2.

CasperWA:
Do you have a use case where we _don't_ want to have the same key/value-pair in the configuration used by the collective data resource strategies (parse, download, resource)? I can't come up with one, and rather like this "one configuration to rule them all"-vibe, in terms of letting a passed configuration represent the whole data resource retrieval - it being information complete.

That said - this is also an issue more related to `oteapi-services` and in particular `otelib`. But the implementation should happen in this package for sure.

For each of your points I don't follow the logic to be honest. Mainly because I don't see the reason for this split.

jesper-friis:
**Conclusion** 
The `ResourceConfig.configuration` is specific to the strategy that the `ResourceConfig` is given to.

So if a parse strategy is calling a download strategy, it is up to the parse strategy properly configure the download strategy.
How this should be done can be seen in the excel_xlsx and (soon) the text_json strategies.

PR #35, opened 2022-01-25
All new tests

TorgeirUstad:
Closes #11 
Add unit tests.

About half of the tests have been written so far.

Note: The JPEG tests fails now, cf. issue #34.

Fixes #24 
Closes #48 
Closes #61

PR #70, opened 2022-02-09
Made datacache accepting AttrDict configuration

jesper-friis:
- Made datacache accepting AttrDict configuration
- Updated the file download strategy according to issue #64 
- Added `read_only=False` as argument to load_workbook() as a workaround for bugs in both openpyxl and Windows


TorgeirUstad (changes requested):
`file.py` doesn't work on Linux, because the host name isn't included in `filename` on line 73 (this is not an issue on Windows).

Se the last commit in PR #35 for a possible workaround.

TorgeirUstad:
Q.E.D.

CasperWA:
> `file.py` doesn't work on Linux, because the host name isn't included in `filename` on line 73 (this is not an issue on Windows).
> 
> Se the last commit in PR #35 for a possible workaround.

This is actually a different issue. See #74

TorgeirUstad:
Ah, good that you are fixing this.

oteapi/datacache/datacache.py
```diff
@@ -17,12 +17,11 @@
 import tempfile
 from contextlib import contextmanager
 from pathlib import Path
-from typing import TYPE_CHECKING
+from typing import TYPE_CHECKING, Mapping
```
CasperWA, line 20:
```suggestion
from typing import TYPE_CHECKING
```

oteapi/datacache/datacache.py
```diff
@@ -95,8 +94,8 @@ def __init__(
     ) -> None:
         if config is None:
             self.config = DataCacheConfig()
-        elif isinstance(config, dict):
-            self.config = DataCacheConfig(**config, extra=Extra.ignore)
+        elif isinstance(config, (Mapping, AttrDict)):
```
CasperWA, line 97:
```suggestion
        elif isinstance(config, (dict, AttrDict)):
```
Not sure `Mapping` is allowed for this (although the tests seem to pass - but if this is the case, we should just use `Mapping` here and make `AttrDict` a subclass of `Mapping` as well as `BaseModel`).

oteapi/strategies/download/file.py
```diff
@@ -30,6 +31,10 @@ class FileConfig(BaseModel):
             "Encoding used when opening the file. The default is platform dependent."
         ),
     )
+    datacache_config: Optional[AttrDict] = Field(
+        AttrDict(),
```
CasperWA, line 35:
```suggestion
    datacache_config: Optional[DataCacheConfig] = Field(
        None,
```
This should use the `DataCacheConfig` model instead of `AttrDict`.
It might also be that instead of having `Optional[DataCacheConfig]` this should simply be `DataCacheConfig` with the default value of `DataCacheConfig()` (instead of the suggested `None` here).

CasperWA (changes requested):

jesper-friis:
Also added default value for `datacache_config` as suggested by Casper

oteapi/strategies/download/file.py
```diff
@@ -31,6 +31,10 @@ class FileConfig(BaseModel):
             "Encoding used when opening the file. The default is platform dependent."
         ),
     )
+    datacache_config: Optional[DataCacheConfig] = Field(
```
CasperWA, line 34:
```suggestion
    datacache_config: DataCacheConfig = Field(
```
As you're providing a default value, which is a default initialization of `DataCacheConfig`, there is no need to allow this field to be `None` by adding `Optional[]`.

CasperWA (changes requested):

CasperWA (approved):
I have merged this with the latest changes implemented in #81.
Note, the main change that's left in the File strategy is changing `cache_config` to `datacache_config` and extending the description slightly.

PR #73, opened 2022-02-10
New Function strategy

### LINKED ISSUE #19: New Function strategy
In order to reduce the amount of boilerplate code that has to be maintained I will suggest that we instead of implementing two new strategies just add a new method to the existing download and parse strategies.

The parse strategies can optionally implement a `serialise()` method that uses `mediaType`/`accessService` from `ResourceConfig` to serialise content into the data cache. 

Similarly can the download strategies optionally implement a `put()` method that uses `downloadUrl`/`accessUrl` from `ResourceConfig` to read content from the cache and upload it. 

The cache is controlled with the same configurations as for downloading and parsing. The `serialisation()` method will, like the `parse()` method, instantiate the download strategy, but call the `put()` method instead.

If implemented like this, the download strategy should at registration time inform whether it implements `get()` and/or `put()`. Alternatively this could be probed automatically by the API. Similarly for the parse strategy.

CasperWA:
Closes #19 

A minimum implementation of a new Function strategy.

There are no specific configuration fields added - other than a `functionType`.

### DIFF
--- a/docs/api_reference/interfaces/ifunctionstrategy.md
+++ b/docs/api_reference/interfaces/ifunctionstrategy.md
@@ -1,0 +1,3 @@
+# ifunctionstrategy
+
+::: oteapi.interfaces.ifunctionstrategy
--- a/docs/api_reference/models/functionconfig.md
+++ b/docs/api_reference/models/functionconfig.md
@@ -1,0 +1,5 @@
+# functionconfig
+
+::: oteapi.models.functionconfig
+    rendering:
+      show_if_no_docstring: true
--- a/oteapi/interfaces/__init__.py
+++ b/oteapi/interfaces/__init__.py
@@ -8,6 +8,7 @@ from typing import Union
 
 from .idownloadstrategy import IDownloadStrategy
 from .ifilterstrategy import IFilterStrategy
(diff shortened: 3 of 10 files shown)


jesper-friis (approved):
Nice clean PR.
Inspired by TransformationConfig, we could consider to add `functionType` and `name` to FunctionConfig. A test implementation would reveal if something more is needed.

CasperWA:
> Nice clean PR. Inspired by TransformationConfig, we could consider to add `functionType` and `name` to FunctionConfig. A test implementation would reveal if something more is needed.

`functionType` is already there, but `name` is not. I thought about that as well, but as these functions will not be long-lived and one wouldn't need to retrieve a name to later on check whether a specific task is done or not, I didn't see the point of having a name - it should only be for logging reasons then, I suppose?

jesper-friis:
> functionType is already there, but name is not. I thought about that as well, but as these functions will not be long-lived and one wouldn't need to retrieve a name to later on check whether a specific task is done or not, I didn't see the point of having a name - it should only be for logging reasons then, I suppose?

Good point, let's keep it as it is now and add more attributes if needed.

PR #78, opened 2022-02-11
Sg/session update model

sygout:
Modification to include SessionUpdate pydantic model.

Branch based on branch with unit tests. No there in the list anymore.

It affects also a few unit tests.


CasperWA (approved):
Nice implementation of having a pydantic model for the return values - cheers @sygout 👍

Issue #71, opened 2022-02-09
Allow the strategies to register the datamodel they expect as configuration


jesper-friis:
Currently there are no way for the user to query the API for what strategy-specific configuration a strategy expects. This is needed.

As a follow up, we should add an endpoint in oteapi-services for querying strategy-specific configurations. Finally, a similar functionality should also be added to otelib.

CasperWA:
The approach for this might be a specific way of handling a new general strategy method in oteapi services.
So if a method like, e.g., `schema`, `dependencies`, `info`, or `introspection` delivers a mapping in the form of a pydantic model, then a services endpoint with an appropriate name `discover`, `info`, etc., can utilize this model to tell the user what is needed from their input.

Or even better, this method is called as part of the `/initialize` endpoint currently in services to understand whether the configuration it got matches the strategy's expected minimum values.
However - thinking more about this, I think that's exactly what `initialization` is for, right? To make sure the configuration is as the strategy would expect and otherwise complain.
What you need more here, is a way for the users to query the OTE API services about what configuration might be needed before they even create a strategy - I don't think this is doable in any way with the current design. From the side of `otelib` this will always be done as a part of running a pipeline. And whether there's a special `introspection` endpoint or it's done as part of `initialize` is irrelevant at that point.
However, it may still be good to "enforce" a strategy to reveal what it expects by adding an extra method to all strategy interfaces that will return a pydantic model, which can be utilized in the `initialize` endpoint of services to validate the passed in configuration and return a meaningful error message to the user.

PR #81, opened 2022-02-14
Pydantic dataclasses

### LINKED ISSUE #63: Image parser: Use datacache
The current image parse strategy takes an image file and creates a cropped file with the same name as the original file, with the prefix "cropped_" added. Both files are stored as local files, but we generally want our resources stored in the datacache instead.

The parser should therefore take a datacache key as input to `get()`, store the cropped file in the datacache, and return the datacache key to the cropped image, like e.g. the download strategies do.

### LINKED ISSUE #68: Remove image/eps as supported image format
The image parse strategy is based on the `Image` class in the `Pillow` package, which supports the `EPS (Encapsulated PostScript)` format. However, reading an `EPS` image requires that `Ghostscript` is installed on the system.

Instead of requiring `Ghostscript` to be installed, we should remove `image/eps` from the list of supported image formats.

### LINKED ISSUE #74: Use proper file scheme URLs in tests
#35 introduced some misuse of `file`-scheme URLs.
A correct `file` scheme url is of the form:
`file:///path/to/file` or `file://hostname/path/to/file`
Where, if `hostname` is `localhost`, the two examples above are equivalent.
The tests use `file://path/to/file`, missing a forward slash and incorrectly using the `host` part of the parsed URL in the `FileStrategy`.

The correct solution here would be to check whether `host` is `None` and use it if not, and otherwise ignore it. This ensures a full handling of `file`-scheme URLs.

### LINKED ISSUE #75: Clean up tests
Static test files, e.g. used for testing strategies, should be placed under the `tests/static/` folder.
Static typing should be fully implemented for all tests.

CasperWA:
Fixes #74 
Closes #75 
Closes #68 
Closes #63 

Use `pydantic.dataclasses.dataclass` instead of `dataclasses.dataclass`.
This ensures the `*_config` attribute is type-checked, meaning one can now sub-class the `*Config` pydantic model and substitute this in for the specific strategy, updating the `configuration` field with another strategy-specific model that then automatically will ensure the configuration for the strategy is correct upon strategy initialization.

Mypy has been extended with the pydantic plugin.

The tests have been slightly streamlined.

This PR also implements the beginnings of satisfying #71. However, it doesn't implement an `introspection` or `schema` method to fully satisfy #71.


CasperWA:
> Will this PR influences OTEAPI-services and OTELIB?

I don't think it should. Everything should be fine from their perspective. But it's difficult to say 100 %. Should be fine though :)

CasperWA:
@jesper-friis I have now updated this PR with what we discussed offline concerning having a utility function for converting URI/IRI/URLs to `pathlib.Path`.

oteapi/strategies/download/sftp.py
```diff
@@ -1,21 +1,19 @@
 """Strategy class for sftp/ftp"""
 # pylint: disable=unused-argument
```
quaat, line 2: better to remove unused arguments than to disable to warning

oteapi/strategies/filter/crop_filter.py
```diff
@@ -1,26 +1,40 @@
...
     from typing import Any, Dict, Optional
 
-    from oteapi.models import FilterConfig
 
+class CropImageConfig(AttrDict):
+    """Configuration model for crop data."""
 
-class SessionUpdateCrop(SessionUpdate):
-    """Class for returning values from crop data."""
+    crop: Tuple[int, int, int, int] = Field(..., description="Box cropping parameters.")
```
quaat, line 17: nice

oteapi/strategies/transformation/celery_remote.py
```diff
@@ -28,11 +38,12 @@ class SessionUpdateCelery(SessionUpdate):
...
 
 
-class CeleryConfig(BaseModel):
-    """Celery configuration."""
+class CeleryStrategyConfig(TransformationConfig):
+    """Celery strategy-specific configuration."""
 
-    task_name: str = Field(..., description="A task name.")
-    args: list = Field(..., description="List of arguments for the task.")
+    configuration: CeleryConfig = Field(
```
quaat, line 44: clever

quaat (approved):

oteapi/strategies/download/sftp.py
CasperWA: Indeed. But I believe this is for the `session` parameter, which I believe we cannot remove? :/

PR #83, opened 2022-02-17
Added support for dumping numpy arrays to the datacache

jesper-friis:


tests/datacache/test_datacache.py
```diff
@@ -34,3 +34,17 @@ def test_cache(tmp_path: "Path") -> None:
     assert len(cache) == 1
     cache.clear()
     assert len(cache) == 0
+
+
+def test_numpy(tmp_path: "Path") -> None:
+    """Test adding numpy arrays in the datachache."""
+    import numpy as np
+
+    from oteapi.datacache.datacache import DataCache
```
francescalb, line 43:
```suggestion
    from oteapi.datacache import DataCache
```
You do not need the extra datacache in the import

francescalb (approved):
Works well, approved, but suggest fixing the extra .datacache in the import.

tests/datacache/test_datacache.py
jesper-friis: Thanks. Your suggestions are included.

In addition we do not check explicit for numpy, but tries to store directly in datacache by default. If it fails we fall back on json serialisation.

jesper-friis: Added an optional `json_encoder` argument to datacache.add(), with which one can supply a json encoder for the data one adds. 

Also added a test that shows how this can be used to add a ase.Atoms object to the datacache.

tests/datacache/test_datacache.py
```diff
@@ -34,3 +36,50 @@ def test_cache(tmp_path: "Path") -> None:
...
+
+def test_ase_atoms(tmp_path: "Path") -> None:
+    """Test adding ase Atoms objects to the datachache."""
+    try:
+        import ase
+    except ImportError:
+        pytest.skip("ase is not installed")
+        return
+
+    from ase.io.jsonio import MyEncoder
```
francescalb: Nice, how did yoy find this? It is not documented as fas as I can see.
https://wiki.fysik.dtu.dk/ase/ase/io/io.html?highlight=io#module-ase.io

PR #102, opened 2022-03-02
Add test for AttrDict.update() method

### LINKED ISSUE #101: Issue with `AttrDict.update()` for `AttrDict` subclasses
It seems there are issues if using `AttrDict` subclasses in the `update()` function.
A test should be written for this and then fixed.

CasperWA:
Fixes #101 

~It seems there is no issue and the issue from which this originates lies elsewhere.~
This simply adds a test for expected valid types to be used with `AttrDict.update()`.

The test has been expanded to also vary the "original" type from which the `update()` method is called.
I.e., the tests now actually tests the interchangeability of types when using the `update()` method on `dict` or `AttrDict` or any sub-class thereof.

The expected error was also found - and fixed by replacing the check for `self.__class__` with `Mapping`.


francescalb (approved):
Looks good

PR #103, opened 2022-03-02
Update "ignore" statements

### LINKED ISSUE #76: Go through ignored dev tools comments
Go through and check validity of ignore statements for development tools like pylint, mypy, bandit, etc.

CasperWA:
Closes #76 

I removed all `# pylint: disable...`, `# type: ignore..` and similar statements and updated the code to either fix the complaints from the various developer tools or reinstate them where there is no fix.

### DIFF
--- a/oteapi/datacache/datacache.py
+++ b/oteapi/datacache/datacache.py
@@ -12,6 +12,7 @@ Features:
 - High performance.
 
 """
+# pylint: disable=too-many-arguments
 import hashlib
 import json
 import tempfile
@@ -141,7 +142,10 @@ class DataCache:
     def add(
         self,
         value: "Any",
-        **options,
+        key: "Optional[str]" = None,
+        expire: "Optional[int]" = None,
+        tag: "Optional[str]" = None,
+        json_encoder: "Optional[Type[json.JSONEncoder]]" = None,
     ) -> str:
         """Add a value to cache.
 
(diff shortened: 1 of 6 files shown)


oteapi/datacache/datacache.py
```diff
@@ -138,13 +138,10 @@ def __del__(self) -> None:
...
 
-    def add(  # pylint: disable=too-many-arguments
+    def add(
         self,
         value: "Any",
-        key: "Optional[str]" = None,
-        expire: "Optional[int]" = None,
-        tag: "Optional[str]" = None,
-        json_encoder: "Optional[Any]" = None,
+        **options,
```
jesper-friis, line 144: What is the benefit replacing explicit keyword arguments with an opaque dict?

oteapi/datacache/datacache.py
```diff
@@ -194,14 +200,11 @@ def get(self, key: str) -> "Any":
...
-    def getfile(  # pylint: disable=too-many-arguments
+    def getfile(
         self,
         key: str,
-        filename: "Optional[Union[Path, str]]" = None,
-        prefix: "Optional[str]" = None,
-        suffix: "Optional[str]" = None,
-        directory: "Optional[str]" = None,
         delete: bool = True,
+        **options,
```
jesper-friis, line 207: Same question as above

oteapi/plugins/entry_points.py
```diff
@@ -459,14 +459,16 @@ def __hash__(self) -> int:
         return hash(tuple(_ for _ in sorted(self._entry_points)))
 
     def __str__(self) -> str:
-        res = {}  # type: ignore[var-annotated]
-        for _ in self._entry_points:
-            if _.type.value in res:
-                res[_.type.value] += 1
+        number_of_strategies: "Dict[str, int]" = {}
+        for entry_point in self._entry_points:
```
jesper-friis, line 463: :+1:

jesper-friis (approved):
Added comments about less explicit keyword arguments

oteapi/datacache/datacache.py
CasperWA: I don't know if there are explicit benefits other than a minimization of the method signature, focusing it in on the most necessary an mostly used parameters.
It's "cleaner", but indeed, a bit more opaque.
This boils down to a "limit" of parameters set by pylint to 5, where these methods were going above 5.

I sort of understand why that limit is set, but at the same time it might be too low.

In the end, I think it's nice to focus the methods in in this way, relegating (un)important method options to a keyword argument.

jesper-friis: I am still not convinced. Hiding keyword arguments in an `**options` makes introspection impossible.
I also thing that the division between important and less important arguments is notion for humans, not machines. Therefore I think this division should be done in the docstring. This is also how numpy does it.

CasperWA: Fair enough. I don't feel strongly in one way or another.

But let me ask you this then: How many parameters are then too many for a function/method? Because then I'll update the pylint configuration with this number - but it should also force up some thoughts around this subject.

CasperWA: I need to push this through now, unfortunately, so I'll make an issue for this, revert the changes specific to this, and redo a disable comment for pylint here.

PR #106, opened 2022-03-03
Update README

### LINKED ISSUE #91: Add some badges to the README
Badges can be sourced from several different sources, e.g., [shields.io](https://shields.io/) or [GitHub itself](https://docs.github.com/en/actions/monitoring-and-troubleshooting-workflows/adding-a-workflow-status-badge).

CasperWA:
Closes #91 

Adds badges, but also extends and updates the README concerning entry points and standard strategies.

Go [here](https://github.com/EMMC-ASBL/oteapi-core/tree/cwa/close-91-update-readme#readme) to see the updates in action on GitHub.
To see the updates in action when built as documentation build and serve the documentation locally (assuming you're in the root of the repository):

```shell
pip install -U -e .[docs]  # not needed if already installed with the `dev` extra
mkdocs serve
```

Now open the URL as returned by the last command - should be something like [http://127.0.0.1:8000/oteapi-core/](http://127.0.0.1:8000/oteapi-core/).

### DIFF
--- a/README.md
+++ b/README.md
@@ -120,9 +120,32 @@ The plugins will then automagically load all installed strategy module plugins, 
 
 OTEAPI Core can be installed with:
 
-```console
-$ pip install oteapi-core
+```shell
+pip install oteapi-core
 ```
+
+### For developers
+
+If you want to install OTEAPI Core to have a developer environment, please clone down the repository from GitHub and install:
+
+```shell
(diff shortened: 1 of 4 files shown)


daniel-sintef (approved):
I think these are very nice and useful updates to the README.

PR #108, opened 2022-03-06
Store image data in datacache instead of session

### LINKED ISSUE #107: The image strategy puts binary data in the session
The image strategy must not place binary data into the session.

Any binary (non-json serialisable and UTF-8 encodable) data placed in the session will make the services fail, due to
"oteapi-services/app/routers/session.py", line 124, in _update_session()
```python
    await redis.set(session_id, session.json().encode("utf-8"))
```

This might be considered a bug in oteapi-services. A workaround is to use the datacache, which is more accepting (supports e.g. the buffer protocol) although it also uses json as serialiser.

jesper-friis:
fixed #107 

Converts image to numpy array and stores it datacache instead of session.
The returned payload is now small and useful.

Additionally:
- cleaned up the image parser and corresponding tests
- provide proper download configurations (via the parser)
- added image_mode option to allow conversion to desired image mode 
- return relevant metadata about the image (not its source), like size, mode, palette, and an info dict
- explicitly closes the Pillow image when done with it, to avoid crash on Windows
- the change in mappingconfig.py is because pydantic does not support typing.Annotated

This PR affects the image parse strategy in oteapi-dlite. Merge oteapi-dlite PR 18 together with this one.

### DIFF
--- a/oteapi/models/mappingconfig.py
+++ b/oteapi/models/mappingconfig.py
@@ -1,11 +1,9 @@
 """Pydantic Mapping Configuration Data Model."""
-from typing import Annotated, Dict, List, Optional
+from typing import Dict, List, Optional, Tuple
 
-from pydantic import Field, conlist
+from pydantic import Field
 
 from oteapi.models.genericconfig import GenericConfig
-
-SemanticTriple = Annotated[list, conlist(str, min_items=3, max_items=3)]
 
 
 class MappingConfig(GenericConfig):
@@ -21,7 +19,7 @@ class MappingConfig(GenericConfig):
             "given as local value/IRI-expansion-pairs."
         ),
     )
-    triples: Optional[List[SemanticTriple]] = Field(
+    triples: Optional[List[Tuple[str, str, str]]] = Field(
         None,
         description="List of semantic triples given as (subject, predicate, object).",
     )
--- a/oteapi/strategies/filter/crop_filter.py
+++ b/oteapi/strategies/filter/crop_filter.py
@@ -45,12 +45,18 @@ class CropImageFilter:
 
     filter_config: CropImageFilterConfig
 
-    def initialize(self, session: "Optional[Dict[str, Any]]" = None) -> SessionUpdate:
+    def initialize(
+        self,
+        session: "Optional[Dict[str, Any]]" = None,
+    ) -> SessionUpdateCropFilter:
         """Initialize strategy and return a dictionary."""
-        return SessionUpdate()
+        return SessionUpdateCropFilter(
+            imagecrop=self.filter_config.configuration.crop,
+        )
 
     def get(
-        self, session: "Optional[Dict[str, Any]]" = None
-    ) -> SessionUpdateCropFilter:
+        self,
+        session: "Optional[Dict[str, Any]]" = None,
+    ) -> SessionUpdate:
         """Execute strategy and return a dictionary"""
-        return SessionUpdateCropFilter(imagecrop=self.filter_config.configuration.crop)
+        return SessionUpdate()
--- a/oteapi/strategies/parse/image.py
+++ b/oteapi/strategies/parse/image.py
@@ -28,8 +28,8 @@ class ImageParserConfig(AttrDict):
         None,
         description="Configuration options for the local data cache.",
     )
-    download_config: Optional[ResourceConfig] = Field(
-        {},
+    download_config: ResourceConfig = Field(
+        ResourceConfig(),
         description="Configurations passed to the downloader.",
     )
     image_key: Optional[str] = Field(
@@ -97,7 +97,7 @@ class SessionUpdateImageParse(SessionUpdate):
         ...,
         description="Image mode. Examples: 'L', 'P', 'RGB', 'RGBA'...",
     )
-    image_palette_key: str = Field(
+    image_palette_key: Optional[str] = Field(
         None,
         description="Datacache key for colour palette if mode is 'P'.",
     )
--- a/oteapi/strategies/parse/image.py
+++ b/oteapi/strategies/parse/image.py
@@ -156,8 +156,9 @@ class ImageDataParseStrategy:
         mime_format = self.parse_config.mediaType.split("/")[1]
         image_format = SupportedFormats[mime_format].value
 
-        # Retrieve image file
+        # Proper download configurations
         conf = self.parse_config.dict()
+        del conf["configuration"]
         if config.download_config:
             conf.update(config.download_config.dict())
         download_config = ResourceConfig(**conf)
--- a/oteapi/strategies/parse/image.py
+++ b/oteapi/strategies/parse/image.py
@@ -1,9 +1,9 @@
 """Strategy class for image/jpg."""
 # pylint: disable=unused-argument
 from enum import Enum
-from io import BytesIO
 from typing import TYPE_CHECKING, Optional, Tuple
 
+import numpy as np
 from PIL import Image
 from pydantic import Field
 from pydantic.dataclasses import dataclass
@@ -22,11 +22,27 @@ class ImageParserConfig(AttrDict):
 
     crop: Optional[Tuple[int, int, int, int]] = Field(
         None,
-        description="Box cropping parameters.",
+        description="Box cropping parameters (left, top, right, bottom).",
     )
     datacache_config: Optional[DataCacheConfig] = Field(
         None,
         description="Configuration options for the local data cache.",
+    )
+    download_config: Optional[ResourceConfig] = Field(
+        {},
+        description="Configurations passed to the downloader.",
+    )
+    image_key: Optional[str] = Field(
+        None,
+        description="Key to use when storing the image data in datacache.",
+    )
+    image_mode: Optional[str] = Field(
+        None,
+        description=(
+            "Pillow mode to convert image into. See "
+            "https://pillow.readthedocs.io/en/stable/handbook/concepts.html "
+            "for details."
+        ),
     )
 
 
@@ -34,7 +50,8 @@ class ImageParserResourceConfig(ResourceConfig):
     """Image parse strategy resource config."""
 
     configuration: ImageParserConfig = Field(
-        ImageParserConfig(), description="Image parse strategy-specific configuration."
+        ImageParserConfig(),
+        description="Image parse strategy-specific configuration.",
     )
     mediaType: str = Field(
         ...,
@@ -48,28 +65,65 @@ class ImageParserResourceConfig(ResourceConfig):
     )
 
 
-class SupportedFormat(Enum):
+class SupportedFormats(Enum):
     """Supported formats for `ImageDataParseStrategy`."""
 
-    JPEG = "JPEG"
-    JPEG2000 = "JPEG2000"
-    PNG = "PNG"
-    GIF = "GIF"
-    TIFF = "TIFF"
-    EPS = "EPS"
+    jpeg = "JPEG"
+    jpg = "JPEG"
+    jp2 = "JPEG2000"
+    png = "PNG"
+    gif = "GIF"
+    tiff = "TIFF"
+    tif = "TIFF"
+    eps = "EPS"
 
 
 class SessionUpdateImageParse(SessionUpdate):
-    """Configuration model for ImageParse."""
+    """Configuration model for ImageParse.
 
-    content: bytes = Field(..., description="Parsed output from ImageParse.")
-    format: SupportedFormat = Field(..., description="The format of the parsed image.")
-    cropped: bool = Field(..., description="Whether or not the image has been cropped.")
+    For
+      - image_mode
+      - image_palette
+      - image_info
+
+    see [Pillow handbook](https://pillow.readthedocs.io/en/stable/handbook/concepts.html) for more details.
+    """
+
+    image_key: str = Field(
+        ...,
+        description="Key with which the image content is stored in datacache.",
+    )
+    image_size: Tuple[int, int] = Field(
+        ...,
+        description="Image size (width, height).",
+    )
+    image_mode: str = Field(
+        ...,
+        description="Image mode. Examples: 'L', 'P', 'RGB', 'RGBA'...",
+    )
+    image_palette_key: str = Field(
+        None,
+        description="Datacache key for colour palette if mode is 'P'.",
+    )
+    image_info: dict = Field(
+        {},
+        description="Additional information about the image.",
+    )
 
 
 @dataclass
 class ImageDataParseStrategy:
     """Parse strategy for images.
+
+    This strategy uses Pillow to read a raw image from the data cache,
+    converts it into a NumPy array and stores the new array in the
+    data cache.
+
+    It also support simple cropping and image conversions.
+
+    The key to the new array and other metadata is stored in the session. See
+    [`SessionUpdateImageParse`][oteapi.strategies.parse.image.SessionUpdateImageParse]
+    for more info.
 
     **Registers strategies**:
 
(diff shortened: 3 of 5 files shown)


oteapi/strategies/filter/crop_filter.py
```diff
@@ -45,12 +45,18 @@ class CropImageFilter:
...
-    def initialize(self, session: "Optional[Dict[str, Any]]" = None) -> SessionUpdate:
+    def initialize(
+        self,
+        session: "Optional[Dict[str, Any]]" = None,
+    ) -> SessionUpdateCropFilter:
         """Initialize strategy and return a dictionary."""
-        return SessionUpdate()
+        return SessionUpdateCropFilter(
+            imagecrop=self.filter_config.configuration.crop,
+        )
```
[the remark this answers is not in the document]
CasperWA: We've agreed offline that `initialize` is the method to be used for filters (and mappings), since this allows them to be used in the most flexible way (both before or after a "source" strategy).

oteapi/strategies/parse/image.py
```diff
@@ -93,53 +147,85 @@ def get(
...
+        downloader = create_strategy("download", download_config)
+        output = downloader.get(session)
+        session.update(output)
+
+        if config.datacache_config and config.datacache_config.accessKey:
+            cache_key = config.datacache_config.accessKey
+        elif "key" in output:
+            cache_key = output["key"]
+        else:
+            RuntimeError("no datacache key provided to downloaded content")
```
CasperWA, line 176:
```suggestion
            RuntimeError("No data cache key provided to the downloaded content")
```

oteapi/strategies/parse/image.py
```diff
@@ -93,53 +147,85 @@ def get(
...
-                    {"version": image.info.get("version", b"")[len(b"GIF") :]}
+            if image.mode == "P":
+                image_palette_key = cache.add(
+                    np.asarray(image.getpalette()), tag=str(id(session))
                 )
+            else:
+                image_palette_key = None
+
+            # The session must be json serialisable - filter out all
+            # non-json serialisable fields in image.info
```
CasperWA, line 210: I think we should consider providing a custom JSON encoder in `oteapi-core` that can be used in `oteapi-services` and similar.

tests/strategies/filter/test_crop_filter.py
```diff
@@ -34,14 +35,16 @@ def test_crop_filter(static_files: "Path") -> None:
     image_parser: "IParseStrategy" = ImageDataParseStrategy(image_config)
 
     # Run "pipeline"
```
CasperWA, line 37:
```suggestion
    # Run pipeline = image_parser > crop_filter
```

tests/strategies/filter/test_crop_filter.py
```diff
@@ -34,14 +35,16 @@ def test_crop_filter(static_files: "Path") -> None:
...
+
+    assert data.shape == (400, 700, 3)
 
-    assert session.get("cropped", False)
-    assert session.get("format", SupportedFormat).value == "JPEG"
-    assert (
-        session.get("content", b"")
-        == (static_files / "sample_700_400.jpeg").read_bytes()
-    )
+    del cache[image_key]
```
CasperWA, line 50: Put this in a `try/finally` block with the `del` line in `finally` and the `data =  ...` and `assert ...` lines are in the `try` block.

tests/strategies/parse/test_image.py
```diff
@@ -70,37 +74,36 @@ def test_image(
...
-
     sample_file = static_files / filename
     reference_file = (
         static_files / filename_cropped if filename_cropped else static_files / filename
     )
     assert sample_file.exists(), f"Test file not found at {sample_file} !"
     assert reference_file.exists(), f"Test file not found at {reference_file} !"
 
-    for config_parse_test in range(2):
-        print(f"config_parse_test={config_parse_test}")
```
CasperWA, line 83: This was quite helpful for debugging and understanding when the test failed why it failed. Please leave in.

tests/strategies/parse/test_image.py
```diff
@@ -70,37 +74,36 @@ def test_image(
...
-            session.get("cropped", False) if crop else not session.get("cropped", False)
-        )
-        assert (
-            session.get("format", SupportedFormat).value
-            == mime_to_format.get(image_format, image_format).upper()
-        )
-        assert session.get("content", b"") == reference_file.read_bytes()
+        session = {"imagecrop": crop} if crop and itest == 1 else {}
+
+        parser = ImageDataParseStrategy(parse_config=config)
```
CasperWA, line 92:
```suggestion
        parser: "IParseStrategy" = ImageDataParseStrategy(parse_config=config)
```

tests/strategies/parse/test_image.py
```diff
@@ -70,37 +74,36 @@ def test_image(
...
+        session.update(parser.initialize(session))
+
+        parser = ImageDataParseStrategy(parse_config=config)
```
CasperWA, line 95:
```suggestion
        parser: "IParseStrategy" = ImageDataParseStrategy(parse_config=config)
```

tests/strategies/parse/test_image.py
```diff
@@ -70,37 +74,36 @@ def test_image(
...
+        session.update(parser.get(session))
+
+        cache = DataCache()
+        image_key = session["image_key"]
+        data = cache.get(image_key)
```
CasperWA, line 100: There should be a `del image_key` here somewhere as well I suppose?
Please put this in another `try/finally` block - essentially we want to make sure the values are always removed from the cache.

CasperWA (changes requested):

oteapi/strategies/parse/image.py
```diff
@@ -48,29 +65,66 @@ class ImageParserResourceConfig(ResourceConfig):
...
+        description="Key with which the image content is stored in datacache.",
+    )
+    image_size: Tuple[int, int] = Field(
+        ...,
+        description="Image size (width, height).",
+    )
+    image_mode: str = Field(
+        ...,
+        description="Image mode. Examples: 'L', 'P', 'RGB', 'RGBA'...",
+    )
```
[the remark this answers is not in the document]
jesper-friis: Yes, this can be inferred from the incoming image. 

But this is a SessionUpdate that is specific for the strategy, so it will only be instantiated in this strategy (or am I wrong?). Since an image always will have a size and mode, I thought it would be a good idea to make that explicit by making these fields non-optional.

oteapi/strategies/parse/image.py
```diff
@@ -93,53 +147,85 @@ def get(
...
-        session = session if session else {}
+        if not session:
+            session = {}
+
+        config = self.parse_config.configuration
+        crop = config.crop if config.crop else session.get("imagecrop")
 
         mime_format = self.parse_config.mediaType.split("/")[1]
-        image_format = self._map_mime_to_format()
+        image_format = SupportedFormats[mime_format].value
```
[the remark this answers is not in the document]
jesper-friis: The mime_format is the key, so It has to be square brackets.

oteapi/strategies/parse/image.py
jesper-friis: Yes, that might be useful. We could do it in the same way as for the data cache...

tests/strategies/filter/test_crop_filter.py
jesper-friis: Yes, maybe we should that, but by default the cache expire within 14 days.

A better cleanup strategy would be to either ask the conftest to set up a new data cache for this testing session that is flushed and deleted when the tests have run.

Another strategy, which would be useful outside testing, is to create a separate Session class (could e.g. be a subclass of AttrDict) with a custom __del__() method that evict all items in the cached that are tagged with the session id `str(id(session))`. That way we can dump temporary data for this session in the cache and be sure it will be cleaned up when the session goes out of scope.

tests/strategies/parse/test_image.py
jesper-friis: ok

tests/strategies/parse/test_image.py
jesper-friis: What is the actual purpose of this cast? Is it to tell that `parser` follows the IParseStrategy protocol? But we already know that parser will have initialize() and get() methods, since it is an instance of ImageDataParseStrategy.

For me it currently makes the code more difficult to read, but if it is useful I have to adapt my eyes...

tests/strategies/filter/test_crop_filter.py
jesper-friis: Added issue #114 for an even better solution, that should be easy to implement. I would prefer that in front of poluting the tests with a lot of `try/finally`...

oteapi/models/mappingconfig.py
```diff
@@ -21,7 +19,7 @@ class MappingConfig(GenericConfig):
             "given as local value/IRI-expansion-pairs."
         ),
     )
-    triples: Optional[List[SemanticTriple]] = Field(
+    triples: Optional[List[Tuple[str, str, str]]] = Field(
```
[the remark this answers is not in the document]
CasperWA: I don't get that locally? Using `typing.Annotated` should exactly fix this issue? Hmm - interesting.

oteapi/strategies/parse/image.py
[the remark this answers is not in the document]
CasperWA: An `Enum` is just a class, you can initialize it with the provided enumeration value like a normal class. I guess one way is as good as the other.

tests/strategies/filter/test_crop_filter.py
CasperWA: I like the idea about a general data cache for all tests, however, this breaks the test compartmentalization a bit. Another solution there is to create a context fixture for the data cache that returns a new data cache and flushes it when exiting.

Concerning a separate `Session` class that flushes the data cache that seems overkill and not the job of session. That should be the job of the data cache or similar.

oteapi/strategies/parse/image.py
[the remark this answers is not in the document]
CasperWA: Well, here it's the configuration you're defining. I.e., this will potentially be the user-supplied information. I don't think it makes sense to require the mode and more here.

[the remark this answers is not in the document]
CasperWA: Ah no - you're right. This is the "output", sorry.

oteapi/strategies/parse/image.py
```diff
@@ -93,53 +143,85 @@ def get(
...
+        crop = config.crop if config.crop else session.get("imagecrop")
 
         mime_format = self.parse_config.mediaType.split("/")[1]
-        image_format = self._map_mime_to_format()
+        image_format = SupportedFormat[mime_format].value
+
+        # Proper download configurations
+        conf = self.parse_config.dict()
+        conf["configuration"] = config.download_config
+        download_config = ResourceConfig(**conf)
```
CasperWA, line 158:
```suggestion
        download_config = self.parse_config.copy()
        download_config.configuration = config.download_config
```

CasperWA: If this doesn't work, you could use `**download_config` in the `create_strategy()` calls below.

oteapi/strategies/parse/image.py
```diff
@@ -93,53 +143,85 @@ def get(
...
+            session_update = SessionUpdateImageParse(
+                image_key=image_key,
+                image_size=image.size,
+                image_mode=image.mode,
+                image_palette_key=image_palette_key,
+                image_info=image_info,
+            )
+
+            # Explicitly close the image to avoid crashes on Windows
+            image.close()
```
CasperWA, line 225: Why not simply return the generated `SessionUpdateImageParse` straight up, instead of storing it in an unused variable first?
```suggestion
            # Explicitly close the image to avoid crashes on Windows
            image.close()

            return SessionUpdateImageParse(
                image_key=image_key,
                image_size=image.size,
                image_mode=image.mode,
                image_palette_key=image_palette_key,
                image_info=image_info,
            )
```

oteapi/strategies/parse/image.py
```diff
@@ -93,53 +143,85 @@ def get(
...
-            self.parse_config.configuration.crop = session["imagecrop"]
-
-    def _map_mime_to_format(self) -> str:
-        """Return mapped Pillow-understandable format."""
-        image_format = self.parse_config.mediaType.split("/")[1]
-        return {
-            "jpg": "JPEG",
-            "jp2": "JPEG2000",
-        }.get(image_format, image_format.upper())
+        return session_update
```
CasperWA, line 227:
```suggestion
```

tests/strategies/filter/test_crop_filter.py
CasperWA: Could you please either put this in a `try/finally` block here - or add a context fixture to do so?

tests/strategies/parse/test_image.py
```diff
@@ -70,37 +74,38 @@ def test_image(
...
         parser: "IParseStrategy" = ImageDataParseStrategy(parse_config=config)
+        session.update(parser.initialize(session))
+
+        parser: "IParseStrategy" = ImageDataParseStrategy(parse_config=config)
+        session.update(parser.get(session))
+
+        cache = DataCache()
+        image_key = session["image_key"]
+        data = cache.get(image_key)
+        del cache[image_key]
```
CasperWA, line 102: Same here, please put this in a `try/finally` block for now.

CasperWA (changes requested):

tests/strategies/filter/test_crop_filter.py
```diff
@@ -34,7 +34,7 @@ def test_crop_filter(static_files: "Path") -> None:
     }
     image_parser: "IParseStrategy" = ImageDataParseStrategy(image_config)
 
-    # Run "pipeline"
+    # Run pipeline = image_parser > crop_filter
```
jesper-friis, line 37:
```suggestion
    # Run pipeline = image_parser >> crop_filter
```

oteapi/models/mappingconfig.py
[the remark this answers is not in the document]
jesper-friis: It could be that I have an older version pydantic. For now I suggest that we go for the list of tuples. We can reintroduce the `SemanticTriple` when we have a requirement of pydantic that we for sure supports it.

oteapi/strategies/parse/image.py
jesper-friis: Added issue 79 to oteapi-services about this.

oteapi/strategies/parse/image.py
jesper-friis: Because we are using the `image` object for creating the `session_update`. But the image must be closed to avoid Windows failing to remove the temporary file as mentioned in the comment.

tests/strategies/filter/test_crop_filter.py
jesper-friis: done

tests/strategies/parse/test_image.py
jesper-friis: ok, done

oteapi/strategies/parse/image.py
```diff
@@ -22,19 +22,36 @@ class ImageParserConfig(AttrDict):
...
         None,
-        description="Box cropping parameters.",
+        description="Box cropping parameters (left, top, right, bottom).",
     )
     datacache_config: Optional[DataCacheConfig] = Field(
         None,
         description="Configuration options for the local data cache.",
     )
+    download_config: Optional[ResourceConfig] = Field(
+        {},
```
CasperWA, line 32: The default value should be either `None` or `ResourceConfig()` - see also the given type for this attribute (`Optional[X]` = `Union[None, X]`).

oteapi/strategies/parse/image.py
```diff
@@ -51,26 +68,59 @@ class ImageParserResourceConfig(ResourceConfig):
...
+    image_size: Tuple[int, int] = Field(
+        ...,
+        description="Image size (width, height).",
+    )
+    image_mode: str = Field(
+        ...,
+        description="Image mode. Examples: 'L', 'P', 'RGB', 'RGBA'...",
+    )
+    image_palette_key: str = Field(
+        None,
```
CasperWA, line 101: The type should either be `Optional[str]` or the default value should be a string - can be an empty string (`""`).

CasperWA (changes requested):
Excellent! Thanks for the changes and patience @jesper-friis. I have two final points that must be addressed, but after this - given that there are no other changes - you can merge once all tests succeed.

oteapi/strategies/parse/image.py
jesper-friis: done

oteapi/strategies/parse/image.py
jesper-friis: ok

PR #110, opened 2022-03-06
Corrected the sql_query_filter.

jesper-friis:
Filters are intended to be placed after the strategy they work on in the pipeline.

Lets agree on the preferred way to set up pipelines. This PR implements the original idea.

### DIFF
--- a/oteapi/strategies/filter/sql_query_filter.py
+++ b/oteapi/strategies/filter/sql_query_filter.py
@@ -35,10 +35,13 @@ class SQLQueryFilter:
 
     filter_config: SqlQueryFilterConfig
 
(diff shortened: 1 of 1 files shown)


CasperWA:
As agreed offline, the strategies should be, to an extent, independent of the order in which they come in the pipeline. At least that should be true for mapping and filter strategies. For other strategies this should perhaps not be true. Hence, we use the `initialize` method for this, as this will make the filter and mapping strategies run _always_ prior to other strategies' `get` methods. The actual order in the pipeline thus becomes unimportant.

CasperWA (approved):

PR #112, opened 2022-03-07
Added mapping strategy

jesper-friis:


oteapi/models/mappingconfig.py
```diff
@@ -1,27 +1,28 @@
...
+        {},
         description=(
             "List of shortnames that expands to an IRI "
             "given as local value/IRI-expansion-pairs."
         ),
     )
-    triples: Optional[List[SemanticTriple]] = Field(
-        None,
-        description="List of semantic triples given as (subject, predicate, object).",
+    triples: Optional[List[Tuple[str, str, str]]] = Field(
```
CasperWA, line 23: I'd rather use `SemanticTriple`, as this can then be reused in other places - please return this to it's original state.

oteapi/models/mappingconfig.py
```diff
@@ -1,27 +1,28 @@
...
     """Mapping Strategy Data Configuration."""
 
     mappingType: str = Field(
-        ..., description="Type of registered mapping strategy. E.g., `mapping/demo`."
+        "mapping",
+        description="Mapping type.  Normally this is just 'mapping'.",
     )
     prefixes: Optional[Dict[str, str]] = Field(
-        None,
+        {},
```
CasperWA, line 17: Either the type should be `Optional[<something>]` and the default value is `None` OR the type is `<something>` and the default value is an "empty"/default <something>.
So this should either stay `None` or the type should have `Optional[]` removed.

oteapi/models/mappingconfig.py
```diff
@@ -1,27 +1,28 @@
...
 
-SemanticTriple = Annotated[list, conlist(str, min_items=3, max_items=3)]
-
 
 class MappingConfig(GenericConfig):
     """Mapping Strategy Data Configuration."""
 
     mappingType: str = Field(
-        ..., description="Type of registered mapping strategy. E.g., `mapping/demo`."
+        "mapping",
```
CasperWA, line 13: Why are you putting in a default here? There should _not_ be a default here.
```suggestion
        ...,
```

oteapi/strategies/mapping/mapping.py
```diff
@@ -0,0 +1,48 @@
...
+    """Strategy for a mapping.
+
+    The mapping strategy simply adds more prefixes and triples to the
+    `prefixes` and `triples` fields in the session.
+
+    Nothing is returned to avoid deleting existing mappings.
+
+    **Registers strategies**:
+
+    - `("filterType", "mapping")`
```
CasperWA, line 24:
```suggestion
    - `("mappingType", "mapping/triples")`
```
Let's call it something more explanatory?

oteapi/strategies/mapping/mapping.py
```diff
@@ -0,0 +1,48 @@
...
+            raise ValueError("Missing session.")
+
+        if not "prefixes" in session:
+            session["prefixes"] = {}
+
+        if not "triples" in session:
+            session["triples"] = []
+
+        session["prefixes"].update(self.config.prefixes)
+        session["triples"].extend(self.config.triples)
```
CasperWA, line 42: So this strategy fails if no `session` is provided and expects the keys `prefixes` and `triples` to exist?

Why not instead create an empty `SessionUpdate()`? Also, you're returning `SessionUpdate()` here, meaning the changes to "session" are not returned from this method.

CasperWA (changes requested):
I think there are some severe shortcuts taken in this PR that doesn't make sense to me.

The `oteapi.models.mappingconfig.py` is the _general_ mapping configuration, not to be confused with an implementation-specific strategy configuration. It should _not_ have default values where they do not make sense.
Rather the strategy implementations should have a mapping default if so desired, or something like this could be implemented in `oteapi-services`, but not here.

There seems to be a `session` dependency in the new mapping strategy that is quite severe. As a session _can_ be `None`/not supplied, this strategy should be more forgiving - creating a new dictionary/`SessionUpdate` if necessary, instead of simply failing.

The chosen mapping type `mapping` is too broad and should be focused in on what is actually being mapped.
If you feel only a single kind of mapping is actually needed, the mapping strategy should be removed in favor of that architecture. Until then, the mapping type should be more explicit.

oteapi/strategies/mapping/mapping.py
CasperWA: Or simply:
```suggestion
    - `("mappingType", "triples")`
```

jesper-friis:
> The `oteapi.models.mappingconfig.py` is the _general_ mapping configuration, not to be confused with an implementation-specific strategy configuration. It should _not_ have default values where they do not make sense. Rather the strategy implementations should have a mapping default if so desired, or something like this could be implemented in `oteapi-services`, but not here.
> 
> The chosen mapping type `mapping` is too broad and should be focused in on what is actually being mapped. If you feel only a single kind of mapping is actually needed, the mapping strategy should be removed in favor of that architecture. Until then, the mapping type should be more explicit.

Yes, the intended use of this mapping strategy is to map datamodel properties to ontological classes. But it should work as well for mapping instance properties to ontological individuals. I have moved the default to the stragegy implementation, but it is not used...

oteapi/models/mappingconfig.py
jesper-friis: Pydantic doesn't support typing.Annotated

oteapi/models/mappingconfig.py
jesper-friis: So to say it is optional (with a sensible default) I should remove `Optional[]` ??

oteapi/strategies/mapping/mapping.py
jesper-friis: fixed

oteapi/models/mappingconfig.py
jesper-friis: I think removing `Optional` is less explicit, but I did as you suggested anyway.

Isn't  `Optional` simply optional in combination with `Field`? If so, I don't understand why we can add it to be explicit.

And that it depends on the value (None/non-None) whether `Optional` should be used or not sounds even more strange. Are you really sure about that?

CasperWA: Well you could use `Optional[]` and still have a default non-`None` value of course - it's just not necessary and simply allows a `None` value for the attribute, which you then have to take into account when using the attribute in your code - hence my original comment. But inherently, you could do as you have done.

CasperWA: If you remove the `Optional[]` and keep a default value for the attribute, you can assume the attribute is _always_  a dictionary and not worry about it being anything else - meaning standard `dict` methods will always work.

oteapi/models/mappingconfig.py
CasperWA: Huh. I had a similar issue with mypy until I used `typing.Annotated` and then it worked? Hmm.

oteapi/models/mappingconfig.py
```diff
@@ -1,27 +1,28 @@
...
             "given as local value/IRI-expansion-pairs."
         ),
     )
-    triples: Optional[List[SemanticTriple]] = Field(
-        None,
-        description="List of semantic triples given as (subject, predicate, object).",
+    triples: Optional[List[Tuple[str, str, str]]] = Field(
+        [],
+        description=(
+            "List of semantic triples given as " "(subject, predicate, object)."
```
CasperWA, line 26:
```suggestion
            "List of semantic triples given as (subject, predicate, object)."
```

oteapi/models/mappingconfig.py
```diff
@@ -1,27 +1,28 @@
...
-SemanticTriple = Annotated[list, conlist(str, min_items=3, max_items=3)]
-
 
 class MappingConfig(GenericConfig):
     """Mapping Strategy Data Configuration."""
 
     mappingType: str = Field(
-        ..., description="Type of registered mapping strategy. E.g., `mapping/demo`."
+        ...,
+        description="Mapping type.  Normally this is just 'triples'.",
```
CasperWA, line 14:
```suggestion
        description="Type of registered mapping strategy.",
```

oteapi/models/mappingconfig.py
```diff
@@ -1,27 +1,28 @@
...
         description=(
             "List of shortnames that expands to an IRI "
             "given as local value/IRI-expansion-pairs."
         ),
     )
-    triples: Optional[List[SemanticTriple]] = Field(
-        None,
-        description="List of semantic triples given as (subject, predicate, object).",
+    triples: Optional[List[Tuple[str, str, str]]] = Field(
+        [],
```
CasperWA, line 24: The same is true here as for the `Optional[dict]` above.

oteapi/strategies/mapping/__init__.py
```diff
@@ -0,0 +1 @@
+from .mapping import MappingStrategy, MappingStrategyConfig
```
CasperWA, line 1:
```suggestion
```
These init files should not import anything. The strategies are retrieved through the `oteapi.plugins` module.

oteapi/strategies/mapping/mapping.py
```diff
@@ -0,0 +1,63 @@
+"""Mapping filter strategy."""
```
CasperWA, line 1:
```suggestion
"""Mapping filter strategy for semantic triples."""
```

oteapi/strategies/mapping/mapping.py
```diff
@@ -0,0 +1,63 @@
...
+    from typing import Any, Dict, Optional
+
+
+class MappingStrategyConfig(MappingConfig):
+    """Strategy-specific mapping config."""
+
+    mappingType: str = Field(
+        "triples",
+        description="Mapping type.  Normally this is just 'mapping'.",
+    )
```
CasperWA, line 19: No need for this - the `mappingType` is set through the `create_strategy()` call, just like any other strategy.
```suggestion
```

oteapi/strategies/mapping/mapping.py
```diff
@@ -0,0 +1,63 @@
...
+
+    Nothing is returned to avoid deleting existing mappings.
+
+    **Registers strategies**:
+
+    - `("mappingType", "triples")`
+
+    """
+
+    config: MappingStrategyConfig
```
CasperWA, line 37:
```suggestion
    config: MappingConfig
```

oteapi/strategies/mapping/mapping.py
```diff
@@ -0,0 +1,63 @@
...
+        if not "triples" in session:
+            session["triples"] = []
+
+        session["prefixes"].update(self.config.prefixes)
+        session["triples"].extend(self.config.triples)
+
+        if return_session:
+            return SessionUpdate(**session)
+        else:
+            return SessionUpdate()
```
CasperWA, line 59: I don't understand your logic at all in this method.
This is what I suggest to have here instead:

```suggestion
        session = session or {}

        if not "prefixes" in session:
            session["prefixes"] = {}

        if not "triples" in session:
            session["triples"] = []

        session["prefixes"].update(self.config.prefixes)
        session["triples"].extend(self.config.triples)

        return SessionUpdate(**session)
```

setup.cfg
```diff
@@ -5,6 +5,8 @@ oteapi.download =
   oteapi.https = oteapi.strategies.download.https:HTTPSStrategy
   oteapi.ftp = oteapi.strategies.download.sftp:SFTPStrategy
   oteapi.sftp = oteapi.strategies.download.sftp:SFTPStrategy
+oteapi.mapping =
+  oteapi.mapping = oteapi.strategies.mapping.mapping:MappingStrategy
```
CasperWA, line 9:
```suggestion
  oteapi.triples = oteapi.strategies.mapping.mapping:MappingTriplesStrategy
```

oteapi/strategies/mapping/mapping.py
```diff
@@ -0,0 +1,63 @@
...
+    """Strategy-specific mapping config."""
+
+    mappingType: str = Field(
+        "triples",
+        description="Mapping type.  Normally this is just 'mapping'.",
+    )
+
+
+@dataclass
+class MappingStrategy:
```
CasperWA, line 23:
```suggestion
class MappingTriplesStrategy:
```

oteapi/strategies/mapping/mapping.py
```diff
@@ -0,0 +1,63 @@
...
+    """Strategy for a mapping.
```
CasperWA, line 24:
```suggestion
    """Strategy for a mapping to semantic triples.
```

oteapi/strategies/mapping/mapping.py
```diff
@@ -0,0 +1,63 @@
...
+
+    The mapping strategy simply adds more prefixes and triples to the
+    `prefixes` and `triples` fields in the session.
+
+    Nothing is returned to avoid deleting existing mappings.
+
```
CasperWA, line 30:
```suggestion
```

tests/strategies/mapping/test_mapping.py
```diff
@@ -0,0 +1,51 @@
+"""Test mapping strategy."""
```
CasperWA, line 1:
```suggestion
"""Test mapping semantic triples strategy."""
```

tests/strategies/mapping/test_mapping.py
```diff
@@ -0,0 +1,51 @@
...
+# pylint: disable=invalid-name
+
+
+def test_mapping() -> None:
+    """Test mapping strategy."""
+    from oteapi.strategies.mapping import MappingStrategy, MappingStrategyConfig
```
CasperWA, line 7:
```suggestion
    from oteapi.models.mappingconfig import MappingConfig
    from oteapi.strategies.mapping import MappingStrategy
```

tests/strategies/mapping/test_mapping.py
```diff
@@ -0,0 +1,51 @@
...
+
+    conf1 = MappingStrategyConfig(
```
CasperWA, line 9:
```suggestion
    conf1 = MappingConfig(
```

tests/strategies/mapping/test_mapping.py
```diff
@@ -0,0 +1,51 @@
...
+        triples=[
+            ("http://onto-ns.com/meta/1.0/Foo#a", "map:mapsTo", "onto:A"),
+            ("http://onto-ns.com/meta/1.0/Foo#b", "map:mapsTo", "onto:B"),
+            ("http://onto-ns.com/meta/1.0/Foo#c", "map:mapsTo", "onto:C"),
+        ],
+    )
+    conf2 = MappingStrategyConfig(
```
CasperWA, line 16:
```suggestion
    conf2 = MappingConfig(
```

CasperWA (changes requested):
Perhaps rename the `mapping.py` strategy file to `triples.py` or `semantic_triples.py` as well or something.

oteapi/models/mappingconfig.py
CasperWA: Please recreate `SemanticTriple` in your image. I.e., `RDFTriple = Tuple[str, str, str]` or similar.

tests/strategies/mapping/test_mapping.py
```diff
@@ -0,0 +1,51 @@
...
+        prefixes={
+            "map": "http://emmo.info/domain-mappings#",
+            "onto": "http://example.com/0.1/Myonto",
```
CasperWA, line 19:
```suggestion
            "onto": "http://example.com/0.1/Myonto#",
```

tests/strategies/mapping/test_mapping.py
jesper-friis: I think this is simply a basic mapping strategy - not specially related to triples.

setup.cfg
jesper-friis: I think I would like to keep the original naming. This is simply a basic mapping strategy that adds mapping triples to the session such that they are available for other strategies.

oteapi/models/mappingconfig.py
jesper-friis: ok

oteapi/models/mappingconfig.py
```diff
@@ -1,25 +1,28 @@
...
             "List of shortnames that expands to an IRI "
             "given as local value/IRI-expansion-pairs."
         ),
     )
-    triples: Optional[List[Tuple[str, str, str]]] = Field(
-        None,
-        description="List of semantic triples given as (subject, predicate, object).",
+    triples: List[RDFTriple] = Field(
+        [],
+        description=("List of RDF triples given as (subject, predicate, object)."),
```
CasperWA, line 27:
```suggestion
        description="List of RDF triples given as (subject, predicate, object).",
```

CasperWA (changes requested):
I don't like that you are changing the general `MappingConfig` (on the same level as `ResourceConfig` and similar). These should _only_ require the very minimum, and then each strategy can make a subclass of them to further increase the requirements, if so desired.

So please return the general `MappingConfig` to _only_ have `mappingType` as a requirement, and please remove any defaults here. This config class is _not_ for a specific strategy's usage, but rather for _any_ given strategy.

Prior to returning `MappingConfig` to it's almost original state, you can copy it over to the `oteapi/strategies/mapping/mapping.py` file, rename it, and have it subclass `MappingConfig`. Here you can ensure that `prefixes` and `triples` are mandatory fields.

CasperWA (approved):
Fantastic! Cheers @jesper-friis

Issue #114, opened 2022-03-07
Allow to bind values added to the data cache to an object, such that they automatically will be remove when the object goes out of scope


jesper-friis:
Could easily be implemented with `weakref.finalize()` by adding a new `bind` argument to DataCache.add()

PR #115, opened 2022-03-07
Implemented #114

jesper-friis:


oteapi/datacache/datacache.py
```diff
@@ -161,6 +163,12 @@ def add(
                 it is taken from the configuration.
             tag: Tag used with [`evict()`][oteapi.datacache.datacache.DataCache.evict]
                 for cleaning up a session.
+            bind: Remove `value` from the cache when the object provided with
+                this argument goes out of scope.  Note that `bind` must be a
+                weakref'able object.
+
+                The argument name refer to that we bind `value` to this
```
francescalb, line 170:
```suggestion
                The argument name refers to that we bind `value` to this
```

francescalb (approved):
Looks good. 
Just a question: As far as I understand if bind is not specified we will end up filling up datachache. Should it maybe be default in some way that we use bind?

jesper-friis:
> Looks good. Just a question: As far as I understand if bind is not specified we will end up filling up datachache. Should it maybe be default in some way that we use bind?

Items we add to the data cache will by default be evicted after 14 days. That default can of cause be adjusted to a more appropriate value

PR #119, opened 2022-03-10
Implement and test `pop()` and `popitem()` for AttrDict

### LINKED ISSUE #118: Add `pop()` (and possibly `popitem()`) to `AttrDict`
These methods are useful when a strategy is called from within another strategy and the session needs to be cleaned up before returning it.

CasperWA:
Closes #118


jesper-friis (approved):
Looks good

PR #124, opened 2022-03-10
Using the `StrategyType` to deliver the `*Config` cls

### LINKED ISSUE #123: Support passing a dictionary as config to `create_strategy()`
It is currently not possible to pass in a dictionary as `config` for `create_strategy()`.
As is the case for `strategy_type`, this should allow a standard Python type as well as the special `*Config` classes.

As this is what would be expected by users, I consider this a bug.

CasperWA:
Fixes #123 

This can then be used to instantiate the correct `config` to use when instantiating a strategy.


oteapi/plugins/entry_points.py
```diff
@@ -118,6 +117,38 @@ def __str__(self) -> str:
     def __repr__(self) -> str:
         return repr(str(self))
 
+    @property
```
jesper-friis, line 120: Move all the imports to the top of the file and simplify the dict.

jesper-friis (approved):
Approved when the comment is addressed.

Otherwise, looks good

PR #125, opened 2022-03-11
Fix pydantic model type and default values

### LINKED ISSUE #117: Fix pydantic model types according to default values
E.g., the `DataCacheConfig` has type `str` but default value `None` for some fields.

All models should be checked and it should be ensured the field types and default values match.

CasperWA:
Fixes #117 

Add the given type for each strategy to make it clearer for each strategy.

### DIFF
--- a/oteapi/models/genericconfig.py
+++ b/oteapi/models/genericconfig.py
@@ -163,7 +163,7 @@ class GenericConfig(BaseModel):
         cls.__fields__["description"].default = cls.__doc__
 
     class Config:
-        """Pydantic configuration for `AttrDict`.
+        """Pydantic configuration for `GenericConfig`.
 
         * **`validate_assignment`**
           Validate and cast set values.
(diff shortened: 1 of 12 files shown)


oteapi/models/genericconfig.py
```diff
@@ -146,3 +161,19 @@ class GenericConfig(BaseModel):
     def __init_subclass__(cls) -> None:
         """Initialize subclass descriptions with their docstrings."""
         cls.__fields__["description"].default = cls.__doc__
+
+    class Config:
+        """Pydantic configuration for `AttrDict`.
```
jesper-friis, line 166:
```suggestion
        """Pydantic configuration for `GenericConfig`.
```

jesper-friis (approved):

jesper-friis (commented):
Approved after fixing the minor typo.

PR #144, opened 2022-03-30
Fix deleting entries in AttrDict

### LINKED ISSUE #143: Setting attributes in `AttrDict` should be handled by pydantic
Currently, a magic method `__setattr__` is in place in `AttrDict` to perform type checking when setting an attribute, however, this should be handled by pydantic instead through the `validate_assignment` configuration.

CasperWA:
# Description:

Fixes #143 

When deleting entries in AttrDict they should be completely removed from the `__dict__` in case the entry is not a defined pydantic field, otherwise it should be reset to the default value and removed from the set of set fields.

Fix `update()` method to check for a pydantic model (after checking for a `dict` or `Mapping`) and then convert the model to a `dict`.

## Type of change:

- [X] Bug fix.
- [ ] New feature.
- [ ] Documentation update.

### DIFF
--- a/oteapi/models/genericconfig.py
+++ b/oteapi/models/genericconfig.py
@@ -4,7 +4,7 @@ from typing import TYPE_CHECKING, Iterable, Mapping
 from pydantic import BaseModel, Field
 from pydantic.fields import Undefined
 
-if TYPE_CHECKING:
+if TYPE_CHECKING:  # pragma: no cover
     from typing import Any, Optional, Tuple, Union
 
 
--- a/oteapi/models/genericconfig.py
+++ b/oteapi/models/genericconfig.py
@@ -17,34 +17,31 @@ class AttrDict(BaseModel, Mapping):
         return self.__dict__.__contains__(name)
 
     def __delitem__(self, key: str) -> None:
-        """Enable deletion access through subscription."""
-        if key in dir(self):
-            self.__delattr__(key)
+        """Enable deletion access through subscription.
+
+        If the item is a pydantic field, reset it and remove it from the set of set
+        fields. Otherwise, delete the attribute.
+
+        """
+        if key in self.__dict__:
             if key in self.__fields__:
-                del self.__fields__[key]
+                # Reset field to default and remove from set of set fields
+                setattr(self, key, self.__fields__[key].default)
                 self.__fields_set__.remove(key)  # pylint: disable=no-member
+            else:
+                self.__delattr__(key)
         else:
             raise KeyError(key)
 
     def __getitem__(self, key: str) -> "Any":
         """Enable read access through subscription."""
-        if key in dir(self):
+        if key in self.__dict__:
             return getattr(self, key)
         raise KeyError(key)
 
-    def __setattr__(self, name: str, value: "Any") -> None:
-        """Extend BaseModel.__setattr__ with type-checking."""
-        if name in self.__dict__ and self.__dict__[name]:
-            target_type = type(self.__dict__[name])
-            if not isinstance(value, target_type):
-                raise TypeError(
-                    "Mapped value must be subclass of " + target_type.__name__
-                )
-        super().__setattr__(name, value)
-
     def __setitem__(self, key: str, value: "Any") -> None:
         """Enable write access through subscription."""
-        self.__setattr__(key, value)
+        setattr(self, key, value)
 
     def __len__(self):
         """Return number of items."""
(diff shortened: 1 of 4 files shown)


tests/models/test_genericconfig.py
```diff
@@ -1,19 +1,46 @@
 """Tests for `oteapi.models.genericconfig`"""
-# pylint: disable=no-member
+# pylint: disable=no-member,pointless-statement,disallowed-name
```
francescalb, line 2: Isn't it better to define pylint disable where they matter rather than as a blanket statement at the top of the file.
I think this way we are teaching ourselves to just disregard pylint when it is inconvenient. I also think this way of doin it increases the risk of adding things that should not be disregarded in a future update.

tests/models/test_genericconfig.py
```diff
@@ -1,19 +1,46 @@
...
+    class CustomConfig(GenericConfig):
+        """A CustomConfig class."""
+
+        configuration: CustomConfiguration
 
 
 @pytest.fixture
-def generic_config() -> "GenericConfig":
+def generic_config() -> "CustomConfig":
     """Return a usable `GenericConfig` for test purposes."""
```
francescalb, line 23:
```suggestion
    """Return a usable `CustomConfig` for test purposes."""
```

tests/models/test_genericconfig.py
```diff
@@ -35,36 +62,92 @@ class SomeSubclass(GenericConfig):
...
+
+def test_attribute_del_item_fail(generic_config: "CustomConfig") -> None:
+    """Ensure KeyError is raised if key does not exist in AttrDict."""
+    non_existent_key = "non_existant_key"
+    assert non_existent_key not in generic_config.configuration
+    with pytest.raises(KeyError):
+        del generic_config.configuration[non_existent_key]
+
+    # Make sure deleting a pydantic field does not fully delete it
+    # and hence does not raise KeyError.
```
francescalb, line 130: Why should this not be fully deleted?

oteapi/models/genericconfig.py
```diff
@@ -17,34 +17,31 @@ def __contains__(self, name: "Any") -> bool:
...
             raise KeyError(key)
 
     def __getitem__(self, key: str) -> "Any":
         """Enable read access through subscription."""
-        if key in dir(self):
+        if key in self.__dict__:
             return getattr(self, key)
         raise KeyError(key)
 
-    def __setattr__(self, name: str, value: "Any") -> None:
```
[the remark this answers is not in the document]
CasperWA: Because it's already implemented in the way we want from the side of pydantic.
One can already do: `my_attr_dict.some_key = some_value` and get type validation should the `some_key` attribute/field/key be a pydantic `Field` defined in some model/class that sub-classes `AttrDict` (like the `SessionUpdate` and other classes).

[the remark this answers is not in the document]
CasperWA: I think the original intent of raising if the type does not match is obstructive. I'd rather have the model cast the value to the type. And note this will again only be an "issue" if the set attribute/field/key is one defined as a pydantic model `Field`.

tests/models/test_genericconfig.py
CasperWA: Yeah that's maybe not clear.
But it should not be fully deleted, because deleting it completely removes the notion of that pydantic attribute/field in the model - that's not something we want. We want the pydantic model to stay intact - instead when deleting a pydantic field, we should reset it to the value it would've had if it was not supplied originally.
At least that's my take on it.

tests/models/test_genericconfig.py
CasperWA: Ah, I missed that this was in the tests, where the main parts of pylint issues don't really apply. However, I can still move this down to be local if you wish?

francescalb: Yes, I still think it should be moved down, so that we do not unintentionally suggest bad practices (to e.g. me when I copy a test to make a new one).

CasperWA: Moved it to the local places now :)

tests/models/test_genericconfig.py
```diff
@@ -35,36 +64,100 @@ class SomeSubclass(GenericConfig):
...
-    try:
-        generic_config.configuration["string"] = 3.14
-        assert False
-    except TypeError:
-        pass
+    generic_config.configuration["string"] = 3.14
+    assert generic_config.configuration["string"] == str(3.14)
 
+    generic_config.configuration["float"] = "0"
+    assert generic_config.configuration["float"] == "0"
```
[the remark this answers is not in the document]
CasperWA: The idea here is to check that if a field/key has not been defined using `pydantic.Field` it will _not_ be cast to any given type. This is equivalent to normal `dict` behaviour, which is what we want :)

PR #178, opened 2022-08-17
Updated sessionupdate datamodel to include session_type and session_id

quaat:
# Description:

## Type of change:

- [ ] Bug fix.
- [ ] New feature.
- [ ] Documentation update.

### DIFF
--- a/oteapi/models/sessionupdate.py
+++ b/oteapi/models/sessionupdate.py
@@ -1,6 +1,41 @@
-"""Pydantic SessionUpdate Data Model."""
-from oteapi.models.genericconfig import AttrDict
+"""Pydantic SessionUpdate Data Model.
 
+Note that the Session Object should be updated according to the
+SessionUpdate. If the Session already has defined a session_type
+and session_id, the code should throw an exception.
+"""
 
-class SessionUpdate(AttrDict):
-    """Session Update Data Model for returning values."""
+from typing import Optional
+from pydantic import ValidationError, root_validator, validator, Field
+from oteapi.models import AttrDict
+
+class OTESession(AttrDict):
+    """ Session Update Data Model for returning data from strategies. """
+    session_type: Optional[str] = Field(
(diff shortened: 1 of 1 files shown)


sygout:
In the pydantic documentation examples, they do not specify:
class Config:
        """ enable assignment validation """
        validate_assignment = True
        
Why is it needed in our case?

CasperWA:
> In the pydantic documentation examples, they do not specify: class Config: """ enable assignment validation """ validate_assignment = True
> 
> Why is it needed in our case?

It validates the field types when the values are updated. Normally, type validation is only done when first creating the model objects.

jesper-friis:
The purpose of introducing `session_type` and `session_id` is unclear from this PR. Is the idea that `session_type` e.g. should be "dlite" or "cuds". What about the `session_id`? For DLite, should that the the id of the underlying collection?

Maybe example use could be added to the description of the pydantic data model.

jesper-friis:
To get rid of the failing tests, install pre-commit and accept the changes suggested by blacken...

oteapi/models/sessionupdate.py
```diff
@@ -1,6 +1,41 @@
...
+and session_id, the code should throw an exception.
+"""
 
-class SessionUpdate(AttrDict):
-    """Session Update Data Model for returning values."""
+from typing import Optional
+from pydantic import ValidationError, root_validator, validator, Field
+from oteapi.models import AttrDict
+
+class OTESession(AttrDict):
```
jesper-friis, line 12: Is this supposed to replace SessionUpdate?

If so, the file should be renamed to otesession.py and all usage should be changed to use SessionUpdate instead. Since, this will be a breaking change, I will suggest that we for backward compatibility keep SessionUpdate a while, but add a deprecation warning.

CasperWA:
I think this should reopened and re-evaluated?

PR #196, opened 2022-10-31
191 psycopg2postgresql datasource strategy

daniel-sintef:
# Description:
Adds a postgres strategy to oteapi. This works but still is too 'pragmatic' and requires some work to make it more in the spirit of oteapi. There are two 'scratch' files that need to be removed before the final merge but are very useful for testing:

1. scratch_launch-docker.sh sets up the postgres docker instance 
2. scratch_test-manually.ipynb sets up a database in our postgres instance and then runs a test manually. 

There are areas I would like some input that are marked explicitly with TODO statements. We should also think a bit about how we can test this after it is pulled, by either (A) building the postgress DB every time in our CI  or (B) with appropriate mocking. 

## Type of change:

- [ ] Bug fix.
- [x] New feature.
- [ ] Documentation update.

### DIFF
--- a/oteapi/strategies/resource/application_vnd_postgres.py
+++ b/oteapi/strategies/resource/application_vnd_postgres.py
@@ -9,38 +9,45 @@ from pydantic.dataclasses import dataclass
 
 from oteapi.datacache import DataCache
 from oteapi.models import AttrDict, DataCacheConfig, ResourceConfig, SessionUpdate
+from oteapi.models.resourceconfig import HostlessAnyUrl
 from oteapi.plugins import create_strategy
 
 if TYPE_CHECKING:  # pragma: no cover
     from typing import Any, Dict
 
 
-# NOTE: common to all SQL configs
-class PostgresParseConfig(AttrDict):
+class PostgresConnectionConfig(AttrDict):
     """Configuration data model for
-    [`PostgresParseStrategy`][oteapi.strategies.parse.application_vnd_postgres.PostgresParseStrategy]."""
+    [`PostgresResourceStrategy`][oteapi.strategies.resource.application_vnd_postgres.PostgresParseStrategy]."""
 
+    # TODO this is *NOT* where the sqlquery should be placed! 
     sqlquery: str = Field("", description="A SQL query string.")
+    # TODO not the final way to pass connection info (should be a dictionary-like object)
+    connection_str: str = Field("", description="String used for connection with postgres DB via psycopg.")
     datacache_config: Optional[DataCacheConfig] = Field(
         None,
         description="Configuration options for the local data cache.",
     )
 
 
-class PostgresParserResourceConfig(ResourceConfig):
-    """Postgres parse strategy resource config."""
+# 
+class PostgresResourceConfig(ResourceConfig):
+    """Postgres resource strategy resource config."""
 
-    mediaType: str = Field(
-        "application/vnd.postgres",
-        const=True,
-        description=ResourceConfig.__fields__["mediaType"].field_info.description,
+    accessUrl: HostlessAnyUrl = Field(
+        ...,
+        description=ResourceConfig.__fields__["accessUrl"].field_info.description,
     )
-    configuration: PostgresParseConfig = Field(
-        PostgresParseConfig(), description="Postgres parse strategy-specific configuration."
+    accessService: Optional[str] = Field(
+        str, description="services to use for access?"
+    )
+    configuration: PostgresConnectionConfig = Field(
+        PostgresConnectionConfig(), description="Postgres resource strategy-specific configuration."
     )
 
 
-def create_connection(conn_config: dict) -> psycopg.Connection:
+# TODO: using a string for conn_config to demo, but should go back to dictionary later
+def create_connection(conn_config: str) -> psycopg.Connection:
     """Create a database connection to Postgres database.
 
     Parameters:
@@ -54,20 +61,20 @@ def create_connection(conn_config: dict) -> psycopg.Connection:
 
     """
     try:
-        return psycopg.connect(**conn_config)
+        return psycopg.connect(conn_config)
     except psycopg.Error as exc:
         raise psycopg.Error("Could not connect to given Postgres DB.") from exc
 
 
-class SessionUpdatePostgresParse(SessionUpdate):
-    """Configuration model for PostgresParse."""
+class SessionUpdatePostgresResource(SessionUpdate):
+    """Configuration model for PostgresResource."""
 
     result: list = Field(..., description="List of results from the query.")
 
 
 @dataclass
-class PostgresParseStrategy:
-    """Parse strategy for Postgres.
+class PostgresResourceStrategy:
+    """Resource strategy for Postgres.
 
     **Registers strategies**:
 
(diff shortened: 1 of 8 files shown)


oteapi/strategies/resource/application_vnd_postgres.py
```diff
@@ -0,0 +1,115 @@
...
+
+if TYPE_CHECKING:  # pragma: no cover
+    from typing import Any, Dict
+
+
+class PostgresConnectionConfig(AttrDict):
+    """Configuration data model for
+    [`PostgresResourceStrategy`][oteapi.strategies.resource.application_vnd_postgres.PostgresParseStrategy]."""
+
+    connection_dict: dict = Field(
```
quaat, line 20: Instead of having a completely generic connection field, it would be better (and more intuitive for the users) if you define all the necessary fields explicitly. e.g. user, password and database

oteapi/strategies/resource/application_vnd_postgres.py
```diff
@@ -0,0 +1,115 @@
...
+    # TODO seems a bit strange to have the sqlquery here, artifact from
+    #     importing from the sqllite version
+    sqlquery: Optional[str] = Field("", description="A SQL query string.")
+
+
+class PostgresResourceConfig(ResourceConfig):
+    """Postgres resource strategy resource config."""
+
+    # TODO: accessUrl and accessService are required, but in this implmentation do nothing, discuss!
+    accessUrl: HostlessAnyUrl = Field(
```
quaat, line 37: accessUrl and accessService is already defined in ResourceConfig, so not necessary to redefine here. accessUrl will upon instantiation name the host of the service (and optionally all other fields such as port, user, password, database)

oteapi/strategies/resource/application_vnd_postgres.py
```diff
@@ -0,0 +1,115 @@
...
+
+    def get(
+        self, session: "Optional[Dict[str, Any]]" = None
+    ) -> SessionUpdatePostgresResource:
+        """Resource Postgres query responses."""
+        if session:
+            self._use_filters(session)
+        session = session if session else {}
+
+        conn_config = self.resource_config.configuration.connection_dict
```
quaat, line 103: Almost, but not quite. The hostname could be fetched from the accessUrl, so ensure that the necessary fields are fetched from the accessUrl first.

scratch_launch-docker.sh
```diff
@@ -0,0 +1 @@
+docker run --name postgres -e POSTGRES_PASSWORD=postgres -d -p 5432:5432  --mount type=bind,source=$(realpath tests),target=/app postgres
```
quaat, line 1: This file can be removed. Better is to add the text in a README.md in a section describing how to launch a postgresql server. Also, the mount-option is rather cumbersome. For an example postgresql-server running in docker you may create a docker volume or just point to some persistent data storage location.
```

docker volume create pgdata
docker run  -d --name postgres \
                   -e POSTGRES_PASSWORD=postgres \
                   -p 5432:5432 \
                   -v pgdata:/var/lib/postgresql/data \
                   --restart always postgres
```

setup.cfg
```diff
@@ -10,6 +10,9 @@ oteapi.mapping =
 oteapi.filter =
   oteapi.filter/crop = oteapi.strategies.filter.crop_filter:CropImageFilter
   oteapi.filter/sql = oteapi.strategies.filter.sql_query_filter:SQLQueryFilter
+oteapi.resource =
+ oteapi.resource/vnd.postgres = oteapi.strategies.resource.application_vnd_postgres:PostgresResourceConfig
```
quaat, line 14: vnd.postgres in not a valid mime-type and doesn't make sence here. Rename to oteapi.resource/postgresql

quaat (changes requested):

oteapi/strategies/resource/postgres.py
```diff
@@ -0,0 +1,176 @@
...
+        params: Optional[str] = "",
+        query: Optional[str] = "",
+        fragment: Optional[str] = "",
+    ):
+
+        """Construct a pydantic AnyUrl based on the given URL properties"""
+
+        # Hostname should always be given
+        if not host:
+            raise Exception("hostname must be specified")
```
jesper-friis, line 56: Use a more specific Exception subclass, like a subclass of ValueError called UrlError.

quaat: Agreed with Jesper.

oteapi/strategies/resource/postgres.py
```diff
@@ -0,0 +1,176 @@
...
+
+        # Update netloc of username or username|password pair is defined
+        netloc = host
+        if user and not password:  # Only username is provided. OK
+            netloc = f"{user}@{host}"
+        elif user and password:  # Username and password is provided. OK
+            netloc = f"{user}:{password}@{host}"
+        else:  # Password and no username is provided. ERROR
+            raise Exception("username not provided")
```
quaat, line 65: Use a more specific Exception subclass (same as above)

oteapi/strategies/resource/postgres.py
```diff
@@ -0,0 +1,176 @@
...
+        """
+
+        # Copy model-state into placeholders
+        config = values.get("configuration")
+        accessUrl = values["accessUrl"]
+
+        # Check and merge user configuration
+        user = accessUrl.user if accessUrl.user else config["user"]
+        if config["user"] and user != config["user"]:
+            raise ValueError("mismatching username in accessUrl and configuration")
```
quaat, line 90: Here ValueError is used, so perhaps use this exception for the previous comments on specific Exception subclass

requirements.txt
```diff
@@ -1,10 +1,12 @@
 agraph-python>=101.0.7,<102
+ase
```
quaat, line 2: Always add the version dependency. The main reason for this is reproducibility.

requirements.txt
```diff
@@ -1,10 +1,12 @@
...
 celery>=5.2.3,<6
 diskcache>=5.4.0,<6
 fastapi-plugins~=0.11.0
 numpy>=1.21,<2
 openpyxl>=3.0.9,<4
 Pillow>=9.0.0,<10
+psycopg
```
quaat, line 9: Add version

quaat (changes requested):
Just some minor changes required

setup.cfg
daniel-sintef: Done

oteapi/strategies/resource/postgres.py
```diff
@@ -0,0 +1,149 @@
...
+                     connection configuration
+
+    Raises:
+        psycopg.Error: If a DB connection cannot be made.
+
+    Returns:
+        Connection object.
+
+    """
+    urlparse_result = urlparse(resource_config.accessUrl)
```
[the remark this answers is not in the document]
daniel-sintef: Done

oteapi/strategies/resource/postgres.py
daniel-sintef: Done

oteapi/strategies/resource/postgres.py
daniel-sintef: Done

requirements.txt
daniel-sintef: Done

requirements.txt
daniel-sintef: Done

oteapi/strategies/resource/postgres.py
daniel-sintef: Done

quaat (approved):
Approved. As mentioned in one comment we should have a follow-up task to clean up the dependencies a bit.

PR #207, opened 2022-11-18
Made AttrDict weakref'able

jesper-friis:
Made AttrDict weakref'able.


oteapi/models/genericconfig.py
```diff
@@ -12,6 +12,12 @@ class AttrDict(BaseModel, Mapping):
     """An object whose attributes can also be accessed through
     subscription, like with a dictionary."""
 
+    # Ensure that AttrDict can be weakref'ed.
+    #
+    # This also exclude creation of the "__dict__" attribute which may save
+    # space and improve spead.
```
CasperWA, line 18:
```suggestion
    # space and improve speed.
```

CasperWA (changes requested):
I am unsure of this change. It seems to destroy the whole purpose of having `AttrDict`, namely to have a dynamic dictionary, at least from [reading the documentation](https://docs.python.org/3/reference/datamodel.html#notes-on-using-slots).
Not to mention `self.__dict__` is referenced several times in this class' methods.

Plus `pydantic` (the parent `BaseModel` class) implements its own `__slots__`, I'm unsure if that fiddles with anything?

jesper-friis:
> I am unsure of this change. It seems to destroy the whole purpose of having `AttrDict`, namely to have a dynamic dictionary, at least from [reading the documentation](https://docs.python.org/3/reference/datamodel.html#notes-on-using-slots). Not to mention `self.__dict__` is referenced several times in this class' methods.

I added `"__dict__"` to `__slots__`. That should fix possible issues.
 
> Plus `pydantic` (the parent `BaseModel` class) implements its own `__slots__`, I'm unsure if that fiddles with anything?

In that case `pydantic` fails to add `"__weakref__"` to `__slots__` which makes AttrDict non-weakable. That is used by oteapi-dlite to ensure that the underlying collection associated to the session lives as long as the session exists. Ok, it is not as simple, because the session is actually created and destroyed for each call the `initialize()` and `get()` strategy methods. However, the two otelib PRs ensure that the get() method of the pipeline return the session. Hence, when the user calls the get() method of a pipeline, a session is first created. Since the session doesn't contain a collection_id, a new collection is created, whos lifetime is bound to the session using [weakref.finalize()](https://docs.python.org/3/library/weakref.html?highlight=weakref#weakref.finalize).  Finally the session is returned to the user. The result is as long the user keep a reference to the session, the collection will stay alive. But when the session is garbage collected, the collection will also be deleted. 

If you want to remove this feature, please provide an alternative implementation to manage the lifetime of resources connected to the session. Also make sure that the SDKdemo continues to work.

quaat:
"Weakref'ability is important for oteapi-dlite to ensure that the underlying collection associated to the session lives as long as the session exists."
I must admit I'm a bit confused here. DLite should really manage its own state and cache without relying on the configurations and session. The session is a means of communicating changes to the specific datamodel configurations of the strategies. DLite might need to save and restore its own state for each strategy if needed.

jesper-friis:
No urgent need for this anymore after PR https://github.com/EMMC-ASBL/oteapi-dlite/pull/84

PR #209, opened 2022-11-19
Updated documentation of prefixes in MappingConfig

### LINKED ISSUE #197: Error in description of MappingConfig.prefixes
The description of the attribute 'prefixes' in MappingConfig describes prefixes as a list of shortnames. This should be changed to 'dictionary of shortnames.

jesper-friis:
Closes #197 

Also avoid dublication of the same triplet.


CasperWA (commented):
What do you say we make the `triples` attribute in the MappingConfig pydantic model a set as well then? Instead of a list?
The cool thing about pydantic models is that it will cast the input to a set, so one can still provide a list as input for `triples` (e.g., through the declarative pipeline), it will just cast the input to a set and store it as such in the attribute.

quaat:
I assume this will not have any effect on the external JSON representation, so I guess this will be fine.

CasperWA:
> I assume this will not have any effect on the external JSON representation, so I guess this will be fine.

None, whatsoever, indeed. It is merely a difference in the internal representation and data validation.

jesper-friis:
> > What do you say we make the `triples` attribute in the MappingConfig pydantic model a set as well then? Instead of a list? The cool thing about pydantic models is that it will cast the input to a set, so one can still provide a list as input for `triples` (e.g., through the declarative pipeline), it will just cast the input to a set and store it as such in the attribute.
> 
> @jesper-friis You have not addressed this review yet.

I have no strong opinions about this as long as it doesn't influence any existing code using the mapping strategy. Feel free to add it.

CasperWA:
> > > What do you say we make the `triples` attribute in the MappingConfig pydantic model a set as well then? Instead of a list? The cool thing about pydantic models is that it will cast the input to a set, so one can still provide a list as input for `triples` (e.g., through the declarative pipeline), it will just cast the input to a set and store it as such in the attribute.
> > 
> > 
> > @jesper-friis You have not addressed this review yet.
> 
> I have no strong opinions about this as long as it doesn't influence any existing code using the mapping strategy. Feel free to add it.

OK. I will add a commit to the PR.

CasperWA (approved):

PR #222, opened 2023-01-10
Enh/secrets json encoders

MBueschelberger:
# Description:

## Type of change:

- [ ] Bug fix.
- [x] New feature.
- [ ] Documentation update.

- [x] Is the change limited to one issue?
- [x] Does this PR close the issue?
- [x] Is the code easy to read and understand?
- [x] Do all new feature have an accompanying new test?

@CasperWA, please see the new `SecretConfig` and the related encoding in the pydantic `Config`, which allows the optional exposing of secrets when it needs to be serialized, e.g. for the RedisCache in ote-services.


oteapi/models/triplestoreconfig.py
```diff
@@ -26,11 +32,19 @@ class TripleStoreConfig(AttrDict):
...
+
+    @root_validator
+    def ensure_user_pass(cls, values: "Dict[str, Any]") -> "Dict[str, Any]":
+        """Ensure that user/password are set, since they are optional in the SecretConfig."""
+        if not all(values.get(_) for _ in ["user", "password"]):
+            raise ValueError("User and password must be defined.")
+        return values
+
+    class Config:
+        """Pydantic configuration for TripleStoreConfig."""
```
CasperWA, line 44: Please add a one or two liner to the doc-string on what `fields` here is doing.

oteapi/models/triplestoreconfig.py
```diff
@@ -1,10 +1,16 @@
...
+from typing import TYPE_CHECKING
 
-from oteapi.models.genericconfig import AttrDict
+from pydantic import Field, root_validator
 
+from oteapi.models.genericconfig import GenericConfig
+from oteapi.models.secretconfig import SecretConfig
 
-class TripleStoreConfig(AttrDict):
+if TYPE_CHECKING:
```
CasperWA, line 9:
```suggestion
if TYPE_CHECKING:  # pragma: no cover
```

oteapi/models/transformationconfig.py
```diff
@@ -29,7 +30,7 @@ class ProcessPriority(str, Enum):
     HIGH = "High"
 
 
-class TransformationConfig(GenericConfig):
+class TransformationConfig(GenericConfig, SecretConfig):  # type: ignore [misc]
```
CasperWA, line 33: I need to check locally why this mypy ignore rule is necessary and whether it can be "fixed". This is true for all of the ignore statements in this case.

oteapi/models/secretconfig.py
```diff
@@ -0,0 +1,67 @@
...
+
+from pydantic import Field, SecretStr
+
+from oteapi.models.genericconfig import AttrDict
+from oteapi.settings import settings
+
+
+def json_dumps(model, *, default):
+    """Alternative function for dumping exposed
+    secrets to json when model is serialized."""
```
CasperWA, line 14:
```suggestion
def json_dumps(model: SecretConfig, *, default: "Callable[[Any], Any]") -> "Dict[str, Any]":
    """Alternative function for dumping exposed
    secrets to json when model is serialized.

    Parameters:
        model: The pydantic model to serialize.
        default: A pass-through to the standard `json.dumps()`'s `default` parameter.
            From the `json.dumps()` doc-string: `default(obj)` is a function that should
            return a serializable version of `obj` or raise `TypeError`.
            The default simply raises `TypeError`.

    Returns:
        The result of `json.dumps()` after handling possible secrets.

    """
```
Two things:

1. Updated the signature to include types.
   For this, some classes needs to be imported from `typing`.
2. Updated the doc-string to be more informative about the parameters.

oteapi/models/secretconfig.py
```diff
@@ -0,0 +1,67 @@
...
+
+        # json_encoders = {
+        #     SecretStr: lambda secret: (
+        #         secret.get_secret_value()
+        #         if settings.expose_secrets and secret
+        #         else secret
+        #     )
+        # }
+        #
+        # Using the json_encoders directly throws an
```
CasperWA, line 57: I'll take a look at this as discussed in a meeting. Feel free to ignore the parts/all of this review until then.

tests/models/test_functionconfig.py
```diff
@@ -0,0 +1,39 @@
+"""Tests for `oteapi.models.functionconfig`"""
+
+
+def test_functionconfig():
+    """Pytest for FunctionConfig, mainly for testing the included secrets."""
+
```
CasperWA, line 6:
```suggestion
```

tests/models/test_triplestoreconfig.py
```diff
@@ -0,0 +1,63 @@
...
+    a configuration object.
+    """
+
+    config = dict(
+        agraphHost="localhost",
+        agraphPort=8080,
+        user="abc",
+        password="pass",
+        repositoryName="test",
+    )
```
CasperWA, line 29:
```suggestion
    config = {
        "agraphHost": "localhost",
        "agraphPort": 8080,
        "user": "abc",
        "password": "pass",
        "repositoryName": "test",
    )
```
It's more efficient to use dict comprehensions (`{}`) to define dicts that instantiating through a call to `dict()`.
Although it's not wrong to do the other thing :)

This suggested change goes for all other instances of `dict()` usages as well.

CasperWA (changes requested):
An initial review. As discussed in the meeting, there are some things I need to test locally and then I'll come back with a full review :)

oteapi/settings.py
```diff
@@ -0,0 +1,23 @@
+"""BaseSettings for oteapi-core"""
```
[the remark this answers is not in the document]
MBueschelberger: @CasperWA, do you think [this ( https://github.com/EMMC-ASBL/oteapi-core/blob/61637dbb47bb0370ddb13c63cf63b69da57a4ffc/oteapi/settings.py#L1 ) is suifficient ? Or should I more closely describe reference the settings-class ?

oteapi/models/triplestoreconfig.py
MBueschelberger: Done 👍

oteapi/settings.py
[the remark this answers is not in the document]
CasperWA: I think that's very sufficient. Thanks! :)

oteapi/models/transformationconfig.py
CasperWA: So this seems to be related to the other unresolved comment here.
I found this rather lengthy, detailed, and interesting discussion on the topic: python/mypy#1237.

As a result, I tried to workaround this "issue" and found a solution I'm happy with:

Don't use `AttrDict` as a base class for `SecretConfig`. There is no need.
Then, additionally, set the `json_dumps` configuration as a class keyword argument (see the 2nd code block in the pydantic documentation [here](https://docs.pydantic.dev/usage/model_config/)).

Now you can remove all the `# type: ignore` lines you've added, also update the type for `model` in the `json_dumps()` method to `dict` and all should be well.
I have done this locally and can push the commit if you want? Otherwise, feel free to do it yourself as well :)

oteapi/models/secretconfig.py
CasperWA: Let's just use `json_dumps`. I think it's elegant enough :)

oteapi/models/triplestoreconfig.py
```diff
@@ -26,11 +32,21 @@ class TripleStoreConfig(AttrDict):
         ...,
         description="AllegroGraph port number.",
     )
-    agraphUser: str = Field(
```
CasperWA, line 29: For backwards compatibility, would it make sense to recreate the `SecretConfig.user` attribute and use the `agraphUser` alias for it? Or maybe it's just silly? I'm not sure how many actually uses it, nor how the attributes here are used elsewhere in general.

oteapi/settings.py
```diff
@@ -0,0 +1,30 @@
+"""BaseSettings for oteapi-core.
+This `configuration/settings`-class is intended to be incorporated as a
+parentclass into the configuration of an FastAPI application.
+See `https://fastapi.tiangolo.com/advanced/settings/` as reference.
+
+Otherwise, check `https://github.com/EMMC-ASBL/oteapi-services/blob/master/app/main.py`
+for a direct example of an inclusion of the OTE api and its settings into an FastAPI instance. """
+from pydantic import BaseSettings, Field
```
CasperWA, line 8:
```suggestion
for a direct example of an inclusion of the OTE api and its settings into an FastAPI instance.
"""
from pydantic import BaseSettings, Field
```

tests/models/test_functionconfig.py
```diff
@@ -0,0 +1,38 @@
+"""Tests for `oteapi.models.functionconfig`"""
+
+
+def test_functionconfig():
```
CasperWA, line 4:
```suggestion
def test_functionconfig() -> None:
```

tests/models/test_triplestoreconfig.py
```diff
@@ -0,0 +1,69 @@
+"""Tests for `oteapi.models.triplestoreconfig`"""
+import pytest
+
+
+def test_triplestoreconfig():
```
CasperWA, line 5:
```suggestion
def test_triplestoreconfig() -> None:
```

tests/models/test_triplestoreconfig.py
```diff
@@ -0,0 +1,69 @@
...
+    """Pytest for TripleStoreConfig."""
+
```
CasperWA, line 7:
```suggestion
```

CasperWA (commented):
Truly minor suggested changes and a comment.
Great work here!

oteapi/models/triplestoreconfig.py
MBueschelberger: I personally wonder why the `TripleStoreConfig` and `Triplestore`-strategy itself are so tightly restricted to agraph. I think it rather makes sense to focus more on the oteapi/tripper interfacing through ontokb in order to make a generic triplestore interface. Once this is provided, this `Triplestore`-strategy might be confusing for a potential oteapi-user/developer in the end.

An alias for `User` in the `TripleStoreConfig`would at least consistency provide a consistency in the attribute-names. I personally think that an `agraph`-prefix make the attributes quite long and it is kind of obvious that agraph is used as backend, when explained in the strategy-description

oteapi/models/secretconfig.py
MBueschelberger: Nice! I was not aware that you can provide kwargs in the class-definition 🤔

oteapi/models/triplestoreconfig.py
CasperWA: I agree with everything here. I'll make an issue to update these classes and to point them towards the new implementations we have.

CasperWA (approved):
Majestic

PR #240, opened 2023-03-02
Polish `celery/remote` transformation strategy

### LINKED ISSUE #236: Ensure celery-strategy is returning correct job-id
This issue goes beyond the oteapi-core. Ensure celery-strategy is returning correct job-id when a remote job is triggered. From the oteapi-service, ensure the status of the job can be correctly introspected. At the termination of the remote celery-worker a return value (model) is submitted back to the broker service (Redis), ensure this data can be fetched from the oteapi-services

CasperWA:
# Description:

Closes #236 

Starting with a test for running the strategy, the related strategy models have been refined. An issue for using the pytest fixtures were found - the Celery documentation has been fixed in celery/celery#7857, but this has yet to be published, which resulted in lots of wasted time :(

Currently, the test is not appropriately testing the strategy, but rather injecting it's own "test" Celery App instead of the strategy's Celery App.
The test should be updated or another should be added to check the actual implementation of the strategy works as intended.

Through this work it has become apparent that an important step is missing: Creating a Celery Task that can actually be referenced in the `get()` method.
I think it makes most sense to register/create tasks in the `initialize()` step of the strategy. Another alternative is to have pre-defined tasks that can be referenced, however, this is slightly opaque for the user.

Note: One must run a local redis instance to test the celery/remote strategy.

## Type of change:

- [x] Bug fix.
- [ ] New feature.
- [ ] Documentation update.

### DIFF
--- a/oteapi/strategies/transformation/celery_remote.py
+++ b/oteapi/strategies/transformation/celery_remote.py
@@ -114,12 +114,18 @@ class CeleryRemoteStrategy:
             session: The current OTE session.
 
         """
+        alias_mapping: dict[str, str] = {
+            field.alias: field_name
+            for field_name, field in CeleryConfig.__fields__.items()
+        }
+
         fields = set(CeleryConfig.__fields__)
         fields |= {_.alias for _ in CeleryConfig.__fields__.values()}
+
         for field in fields:
             if field in session:
                 setattr(
                     self.transformation_config.configuration,
-                    field,
+                    alias_mapping[field],
                     session[field],
                 )
--- a/oteapi/strategies/transformation/celery_remote.py
+++ b/oteapi/strategies/transformation/celery_remote.py
@@ -28,15 +28,24 @@ CELERY_APP = Celery(
 )
 
 
-class CeleryConfig(AttrDict):
+class CeleryConfig(AttrDict, allow_population_by_field_name=True):
     """Celery configuration.
 
     All fields here (including those added from the session through the `get()` method,
     as well as those added "anonymously") will be used as keyword arguments to the
     `send_task()` method for the Celery App.
+
+    Note:
+        Using `alias` for the `name` field to favor populating it with `task_name`
+        arguments, since this is the "original" field name. I.e., this is done for
+        backwards compatibility.
+
+    Setting `allow_population_by_field_name=True` as pydantic model configuration in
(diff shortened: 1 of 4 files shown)


CasperWA:
Great success concerning the new test and running it in CI: https://github.com/EMMC-ASBL/oteapi-core/actions/runs/4316062964/jobs/7531325447#step:8:89
Only thing is that we will have to split up the pytest CI job depending on OS as the Windows OS does not support using the `services` key.

oteapi/strategies/transformation/celery_remote.py
```diff
@@ -30,19 +29,21 @@
...
-    """Celery configuration."""
+    """Celery configuration.
 
-    task_name: str = Field(..., description="A task name.")
+    All fields here (including those added from the session through the `get()` method,
+    as well as those added "anonymously") will be used as keyword arguments to the
+    `send_task()` method for the Celery App.
+    """
+
+    name: str = Field(..., description="A task name.")
```
quaat, line 39: Note that this will break the current demonstrators which use task_name in the configuration.

oteapi/strategies/transformation/celery_remote.py
```diff
@@ -74,21 +75,16 @@ class CeleryRemoteStrategy:
 
     def get(self, session: "Optional[Dict[str, Any]]" = None) -> SessionUpdateCelery:
         """Run a job, return a job ID."""
-        celery_kwargs = {}
         if session:
             self._use_session(session)
-            celery_kwargs = session.copy()
```
quaat, line 80: We need to use the valid fields of the session as the configuration. I.e. the specific configuration might come from the conf-object directly OR through the session.

quaat (changes requested):

oteapi/strategies/transformation/celery_remote.py
CasperWA: I understand. But this is the correct argument name for the `send_task()` method. I could get around it by using `alias`?

CasperWA: I have implemented using `alias`, as well as ensuring non-aliases can be used to instantiate the strategy-specific configuration model (`CeleryConfig`). Furthermore, I've expanded the `_use_session()` method to include the possibility of using aliases in the session object.

quaat (approved):

PR #307, opened 2023-08-23
Update .pre-commit-config.yaml

Treesarj:
# Description:
change pydantic version, to remove the mypy errors caused from latest pydantic versions- > 2

## Type of change:

- [ ] Bug fix.
- [ ] New feature.
- [ ] Documentation update.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -45,7 +45,7 @@ repos:
       exclude: ^tests/.*$
       additional_dependencies:
       - "types-requests"
-      - "pydantic"
+      - "pydantic>=1.9.0,<2"
 
   - repo: https://github.com/SINTEF/ci-cd
     rev: v2.4.0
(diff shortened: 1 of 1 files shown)


.pre-commit-config.yaml
```diff
@@ -45,7 +45,7 @@ repos:
       exclude: ^tests/.*$
       additional_dependencies:
       - "types-requests"
-      - "pydantic"
+      - "pydantic>=1.9.0,<2"
```
CasperWA, line 48:
```suggestion
      - "pydantic<2"
```
I think, since we don't want to keep updating the minimum version, keeping it similar to the requirements file, I'd just put less than 2 here instead. This will ensure it will be the latest v1 minor and patch version that is installed, always.

Simultaneously, an issue should be raised to update pydantic support in the code base to v2, removing the version restriction from this line (or maybe setting it to `>=2` instead?)

CasperWA (approved):

PR #318, opened 2023-09-01
Added support for relative file paths

jesper-friis:
# Description:
Even though that OTEAPI is designed for working on remote data sources, local files are still a common use case. Currently the file strategy only accepts absolute paths, which is an annoying complication in demonstrations that should work for more than one person. 

The fix is to relay on urllib instead of the buggy FileUrl implementation in pydantic.

This PR only involves a small change in `oteapi/utils/paths.py`. All the other files are dragged in by pre-commit...

## Type of change:

- [x] Bug fix.
- [ ] New feature.
- [ ] Documentation update.

### DIFF
--- a/oteapi/utils/paths.py
+++ b/oteapi/utils/paths.py
@@ -31,8 +31,10 @@ def uri_to_path(uri: "Union[str, AnyUrl, ParseResult]") -> Path:
         A properly converted URI/IRI/URL to `pathlib.Path`.
 
     """
-    uri = urlparse(uri)  # type: ignore
-    uri_path = uri.netloc + uri.path
+    if not isinstance(uri, ParseResult):
+        uri = urlparse(uri)  # type: ignore
+
+    uri_path = uri.netloc + uri.path if uri.scheme == "file" else uri.path
 
     if uri.scheme != "file":
         warnings.warn(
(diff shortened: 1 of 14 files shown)


oteapi/utils/paths.py
```diff
@@ -31,13 +31,10 @@ def uri_to_path(uri: "Union[str, AnyUrl, ParseResult]") -> Path:
...
-        uri_path = uri.path
-    elif isinstance(uri, str):
-        uri = urlparse(uri)
-        uri_path = uri.path
-    else:
-        raise TypeError("uri is expected to be either a string or parsed URI/IRI/URL.")
+    if not isinstance(uri, ParseResult):
+        uri = urlparse(uri)  # type: ignore
+
+    uri_path = uri.netloc + uri.path if uri.scheme == "file" else uri.path
```
quaat, line 37: To avoid precedence problems (or confusing readability), please use parentheses to clarify it the `if` statement should include the expression `uri.netloc + uri.path` or if you are just checking the `uri.path`

quaat (changes requested):

oteapi/utils/paths.py
jesper-friis: done

quaat (approved):

quaat (approved):

PR #330, opened 2023-09-19
Upgrade to pydantic v2

### LINKED ISSUE #308: Make the code-base compatible with pydantic version 2 and above

CasperWA:
# Description:

Upgrade to using pydantic v2, dropping support for v1.

I originally intended for this change to incorporate support for both v1 and v2, but I don't see the point.
~Indeed, v1 is (for now) kept in the pydantic library under a special `v1` module - so another way of "upgrading" to support pydantic v2 would be to simply add this to the import statements from pydantic.~
**EDIT**: Did this, and it did not work as intended. A split in the development is now underway, see #346.
But using the new classes and updated system utilizes the underlying rust implementations for effeciency improvements, as well as ensures updated support for future v2 releases.

One thing I've come across now is pylint's inability to understand pydantic models' field values.
~I suggest to switch out pylint usage with [ruff](https://docs.astral.sh/ruff/) instead.~ It is much faster, supports pydantic v2, but does not have the complete ruleset that exists in pylint, however. Using mypy helps a lot, though. For more information see [here](https://docs.astral.sh/ruff/faq/#how-does-ruff-compare-to-pylint).
**EDIT**: Has been implemented by #332 (issue: #331).

Something else I noticed during this upgrade, is that one can no longer instantiate a parent config class, e.g., `ResourceConfig` and use that class instance/object for any strategy requiring (originally) a `ResourceConfig`, but has since created a strategy-specific model, subclassing `ResourceConfig`.
In terms of specificity of objects, this sort of makes sense. It means that one now has to _always_ pass in a Python dictionary, which will then instantiate the correct config model class as well as perform any validation down the line. At least that should be the preferred and recommended way of doing this.

## Type of change:

- [ ] Bug fix.
- [x] New feature.
- [ ] Documentation update.


CasperWA:
I'm getting second thoughts about completely dropping pydantic v1 support. But I'm afraid supporting both major versions will be an implementation nightmare with an extreme amount of conditional statements, which are essentially throw-away work, since once the majority of libraries have implemented support for pydantic v2, then all of that work should be reverted.

CasperWA:
@daniel-sintef Would it be possible for you to review the parts affecting the postgres parser strategy? I have implemented quite a few changes here to support pydantic v2 as well as trying to "simplify" and minimize the code a bit.

CasperWA:
@Treesarj I'll leave the major over-all review to you :)

oteapi/strategies/parse/postgres.py
```diff
@@ -35,78 +33,78 @@ class PostgresResourceConfig(ResourceConfig):
         description="Configuration options for the local data cache.",
     )
 
+    @model_validator(mode="before")
     @classmethod
-    def _urlconstruct(
```
daniel-sintef, line 39: Why was this section removed? @quaat  was a key architect of this section so you should check a bit with him before removing

oteapi/strategies/parse/postgres.py
```diff
@@ -35,78 +33,78 @@ class PostgresResourceConfig(ResourceConfig):
...
+        accessUrl = AnyUrl(data["accessUrl"])
+        default_config = PostgresConfig()
+
+        if not accessUrl.host:
+            raise ValueError("missing host in accessUrl")
 
         # Check and merge user configuration
-        user = accessUrl.user if accessUrl.user else config["user"]
-        if config["user"] and user != config["user"]:
+        user = (
```
daniel-sintef, line 60: the method seems very hard to read and verbose. I'm not very clear why this has been changed, is it for compatibility reasons?

oteapi/strategies/parse/postgres.py
CasperWA: Because the construction is now implemented as a part of the type. This can be done inherently with pydantic and utilizing the `typing.Annotated` class.
In any case, I'll add @quaat as a reviewer of the changes in this file as well 👍

oteapi/strategies/parse/postgres.py
CasperWA: I suppose a reversion to using placeholder variables can be done for verbosity's and readability's sake.

CasperWA: Done

oteapi/strategies/parse/postgres.py
CasperWA: So just an additional comment here. The construction of the URL can be done with the `AnyUrl`'s `build()` class method. This was also the case in pydantic v1.
For pydantic v2 one can go further and add any constraints to a url by using `typing.Annotated` and pydantic v2's new `Url` and `UrlConstraints` classes. Furthermore, a `PostgresUrl` class already existed in pydantic's arsenal. But I suppose there was a desire to support `http`/`https` url schemes in this strategy as well, hence it was not used? In any case, this is easily done now - however, the specific checks and double checks done for each part of the URL that is being done here, is not possible with this - only data type specification and validation is possible, hence this approach is not used for `accessUrl` in this case (but it is used now for other strategies to replace subclassing `AnyUrl`).

oteapi/strategies/parse/postgres.py
daniel-sintef: it still seems like we can maybe refactor a bit more, how do you feel about something like: 

```python
if "accessUrl" not in data:
    raise ValueError("missing accessUrl in PostgreSQL resource configuration")

# Copy model-state into placeholders
accessUrl = AnyUrl(data["accessUrl"])
default_config = PostgresConfig()
current_config: dict[str, Any] = data.get("configuration", {})

if not accessUrl.host:
    raise ValueError("missing host in accessUrl")

def _get_and_validate_config_value(attr, config_key):
    """Get value from accessUrl or current_config, and check for mismatches."""
    value_from_url = getattr(accessUrl, attr)
    value_from_config = current_config.get(config_key, getattr(default_config, config_key))
    
    final_value = value_from_url or value_from_config
    
    if value_from_config and final_value != value_from_config:
        raise ValueError(f"mismatching {config_key} in accessUrl and configuration")
    
    return final_value

user = _get_and_validate_config_value("username", "user")
password = _get_and_validate_config_value("password", "password")
dbname = _get_and_validate_config_value("path", "dbname")
```

daniel-sintef (approved):
I'm approving now since I might not have a lot of time later. I think I've made all the major points I wanted to make. Do add the helper function I suggested unless the other reviewers have objections

CasperWA:
> (...) Do add the helper function I suggested unless the other reviewers have objections

If you meant [this](https://github.com/EMMC-ASBL/oteapi-core/pull/330#discussion_r1345477962), then it's already added in bcb7101.

oteapi/models/genericconfig.py
```diff
@@ -1,104 +1,125 @@
...
+    model_config = ConfigDict(
+        extra="allow", validate_assignment=True, arbitrary_types_allowed=True
+    )
 
-    def __getitem__(self, key: str) -> "Any":
-        """Enable read access through subscription."""
-        if key in self.__dict__:
+    # Collection methods
+    def __contains__(self, key: object) -> bool:
+        """Mapping `__contains__`-method."""
```
daniel-sintef, line 21: Go over the doc strings to make sure you're referencing the correct class, e.g. here it should be 'Collection'

oteapi/models/genericconfig.py
```diff
@@ -113,36 +134,16 @@ def popitem(self) -> "Tuple[str, Any]":
...
         """
         if not self:
             raise KeyError(f"popitem(): {self.__class__.__name__} is empty")
 
-        key = list(self.__dict__)[-1]
+        key = list(self.keys())[-1]
         value = self.pop(key)
         return key, value
 
-    class Config:
```
daniel-sintef, line 126: make sure these doc strings are transferred to where they occur in the new version

tests/models/test_genericconfig.py
```diff
@@ -220,7 +211,7 @@ def test_attrdict_pop_popitem() -> None:
     assert len(attrdict) == len(data) - 1
 
     # Should follow LIFO (last-in, first-out) wrt `__dict__`
```
daniel-sintef, line 213: comment should be fixed

oteapi/models/genericconfig.py
```diff
@@ -1,104 +1,125 @@
...
 
     def items(self):
         """Return a view of all (key, value) pairs."""
-        return self.__dict__.items()
+        return self.model_dump().items()
 
     def keys(self):
         """Return a view of all keys."""
-        return self.__dict__.keys()
+        return self.model_dump().keys()
```
daniel-sintef, line 45: give a final google-search check for speed implications

oteapi/models/genericconfig.py
```diff
@@ -1,104 +1,125 @@
...
-        """Implement the != operator."""
-        if isinstance(other, BaseModel):
-            return self.dict() != other.dict()
-        return self.dict() != other
-
-    def update(
-        self, other: "Optional[Union[dict, AttrDict, Iterable[Any]]]" = None, **kwargs
+        return getattr(self, key, default)
+
+    def __eq__(self, value: object) -> bool:
```
daniel-sintef, line 55: doc string?

daniel-sintef: or remove docstrings elsewhere? it should be consistent either way

oteapi/models/genericconfig.py
```diff
@@ -1,104 +1,125 @@
 """Generic data model for configuration attributes."""
-from typing import TYPE_CHECKING, Iterable, Mapping
+from typing import TYPE_CHECKING, Iterable, Mapping, MutableMapping
```
daniel-sintef, line 2: Iterable, Mapping and MutableMapping are deprecated in typing and should now be imported from collections.abc

oteapi/models/genericconfig.py
```diff
@@ -159,22 +160,8 @@ class GenericConfig(BaseModel):
     )
 
     @classmethod
-    def __init_subclass__(cls) -> None:
+    def __init_subclass__(cls, **kwargs) -> None:
         """Initialize subclass descriptions with their docstrings."""
-        cls.__fields__["description"].default = cls.__doc__
-
-    class Config:
```
daniel-sintef, line 166: check if info in this doc string needs to be moved elsewhere

oteapi/models/secretconfig.py
```diff
@@ -1,44 +1,12 @@
...
-from pydantic import BaseModel, Field, SecretStr
+from pydantic import BaseModel, Field, SecretStr, field_serializer
 
 from oteapi.settings import settings
 
-if TYPE_CHECKING:  # pragma: no cover
-    from typing import Any, Callable
 
-
-def json_dumps(model: dict, default: "Callable[[Any], Any]") -> "str":
```
daniel-sintef, line 13: Check if external people are using this, consider aliasing this

oteapi/plugins/factories.py
```diff
@@ -76,11 +76,14 @@ def make_strategy(
             )
 
         if isinstance(config, dict):
```
daniel-sintef, line 78: maybe add a comment to describe what is happening here a bit more

oteapi/strategies/download/file.py
```diff
@@ -47,7 +47,8 @@ class FileResourceConfig(ResourceConfig):
         FileConfig(), description="File download strategy-specific configuration."
     )
 
-    @validator("downloadUrl")
+    @field_validator("downloadUrl")
+    @classmethod
```
daniel-sintef, line 51: 👍

oteapi/strategies/parse/text_csv.py
```diff
@@ -233,10 +234,9 @@ class CSVConfig(AttrDict):
 class CSVResourceConfig(ResourceConfig):
     """CSV parse strategy filter config."""
 
-    mediaType: str = Field(
+    mediaType: Literal["text/csv"] = Field(
         "text/csv",
```
daniel-sintef, line 238: Consider comment: "text/csv", # this should match the string in the Literal above

daniel-sintef: I don't like the way you have to write out "text/csv" twice, sounds like you need to do this, but it seems weird

oteapi/utils/paths.py
```diff
@@ -4,11 +4,11 @@
 from typing import TYPE_CHECKING
 from urllib.parse import ParseResult, urlparse
 
-from pydantic import AnyUrl
-
 if TYPE_CHECKING:  # pragma: no cover
     from typing import Union
```
daniel-sintef, line 9: empty space

pyproject.toml
```diff
@@ -34,7 +35,6 @@ dependencies = [
     "psycopg~=3.1.10",
     "pysftp~=0.2.9",
     "requests>=2.26.0,<3",
```
daniel-sintef, line 37: consider placing this in a separate PR, removing this requirement is not linked to the pydantic update

tests/models/test_functionconfig.py
```diff
@@ -32,7 +32,11 @@ def test_functionconfig() -> None:
     }
 
     settings.expose_secrets = False
-    assert FunctionConfig(**base_config).json() == json.dumps(config_hidden)
+    assert FunctionConfig(**base_config).model_dump_json() == json.dumps(
+        config_hidden, indent=None, separators=(",", ":")
+    )
```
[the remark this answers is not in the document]
daniel-sintef: NOTE: We should consider adding docs for plugin developers (but that is a whole other issue/PR!)

tests/models/test_genericconfig.py
```diff
@@ -84,11 +82,19 @@ def test_attribute_set(generic_config: "CustomConfig") -> None:
...
+    One should instead use Annotated to create custom types.
+    See https://docs.pydantic.dev/dev-v2/usage/types/custom/ for more information.
 
     """
+    from pydantic import ValidationError
+
     generic_config.configuration["string"] = "bar"
-    generic_config.configuration["string"] = 3.14
-    assert generic_config.configuration["string"] == str(3.14)
+    # Will not work with Pydantic v2:
```
daniel-sintef, line 93: this comment should be removed, or at least moved to the note at the beginning of the test

tests/models/test_genericconfig.py
```diff
@@ -105,38 +111,26 @@ def test_attribute_del_item(generic_config: "CustomConfig") -> None:
...
 
     # "string" is defined as a pydantic model field.
-    # This means it should not be deleted, but reset to its default value
-    # and removed from the set of set fields.
+    # This means it should not be deleted from the model schema,
+    # but it should be removed from the model isntance.
     generic_config.configuration["string"] = "my secret string"
     assert generic_config.configuration.string == "my secret string"
-    assert "string" in generic_config.configuration.__fields_set__
+    assert "string" in generic_config.configuration.model_fields_set
```
daniel-sintef, line 119: The behavior of deleting used to be resetting the field. And there were a lot of tests using the logic of reset. Consider adding the reset and then preserving the tests, and also adding a deprecation warning in the delete function

tests/models/test_triplestoreconfig.py
```diff
@@ -58,10 +58,14 @@ def test_triplestoreconfig() -> None:
     }
 
     settings.expose_secrets = False
-    assert TripleStoreConfig(**config).json() == json.dumps(config_hidden)
+    assert TripleStoreConfig(**config).model_dump_json(
```
daniel-sintef, line 61: consider making a comment why token, client_id and client_secret must be excluded

daniel-sintef:
Please add in tests for values, clear & update for the generic config

oteapi/models/secretconfig.py
CasperWA: As far as I know, no external people are calling or using this function directly.

oteapi/strategies/parse/text_csv.py
CasperWA: The type explains that the value can _only_ be `"text/csv"`, which means validation will fail if it is anything else. This is the behavior we want for the CSVStrategy.
The second `"text/csv"` is the default value that will be set for `mediaType` should it not be explicitly given.

One could argue that `mediaType` must ALWAYS be explicitly provided, and hence the default value can be changed to an ellipsis (`...`) to signify this? I don't mind that change at all. This will reflect what is originally intended for the `ResourceConfig` as well, but can only be realized through a custom validator due to it only being valid if different pairs of fields are given at the same time (`mediaType` + `downloadUrl` OR `accessUrl` + `accessService`).

tests/models/test_functionconfig.py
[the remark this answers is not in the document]
CasperWA: I think I'll add a comment here instead.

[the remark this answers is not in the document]
CasperWA: Also, an issue has been opened for tracking plugin developer docs: #355

tests/models/test_triplestoreconfig.py
CasperWA: I've updated this logic to again by default exclude these fields when serializing the `TripleStoreConfig` model - so any changes here can be reverted (except those related to changes to `json.dumps()`.

tests/models/test_genericconfig.py
```diff
@@ -105,38 +111,26 @@ def test_attribute_del_item(generic_config: "CustomConfig") -> None:
...
 
     del generic_config.configuration["string"]
-    assert "string" in generic_config.configuration
-    assert (
-        generic_config.configuration.string
-        == generic_config.configuration.__fields__["string"].default
-    )
-    assert "string" not in generic_config.configuration.__fields_set__
+    assert "string" not in generic_config.configuration
+    assert "string" in generic_config.configuration.model_json_schema()["properties"]
```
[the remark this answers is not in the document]
CasperWA: This part has been completely updated, including updated comments.

.github/workflows/ci_check_dependencies.yml
```diff
@@ -47,14 +47,16 @@ jobs:
...
+      # Follow issue EMMC-ASBL/oteapi-services#328 for more information about the
+      # urllib3 dependency.
       ignore: |
         dependency-name=agraph-python...version=>=103
         dependency-name=celery...version=>=6
         dependency-name=diskcache...version=>=6
         dependency-name=numpy...version=>=2
         dependency-name=openpyxl...version=>=4
         dependency-name=Pillow...version=>=11
-        dependency-name=pydantic...version=>=2
```
daniel-sintef, line 57: version 3?

CasperWA: Yeah. This job is for the "standard" pydantic v2 development branch. So we up the ignore rule to ignore pydantic v3. We _want_ updates to new v2 versions :)

oteapi/plugins/factories.py
```diff
@@ -75,12 +75,20 @@ def make_strategy(
                 "strategy_type should be either of type StrategyType or a string."
             )
 
+        # 'config' must be a dict when instantiating the strategy's implementation.
```
daniel-sintef, line 78: suggested change: 

# The 'strategy_type' should be of the 'StrategyType' type or a string.
# When initializing the strategy's implementation, 'config' must be a dictionary.
# The 'config_model' fetches the desired strategy.
# By establishing 'config_model', we ensure the 'config' matches the strategy type,
# minimizing the chances of logic errors.
if isinstance(config, dict):

oteapi/models/triplestoreconfig.py
```diff
@@ -1,13 +1,13 @@
...
+from pydantic import Field, model_validator
 
 from oteapi.models.genericconfig import GenericConfig
-from oteapi.models.secretconfig import SecretConfig
+from oteapi.models.secretconfig import SecretConfig, TogglableSecretStr
 
-if TYPE_CHECKING:  # pragma: no cover
-    from typing import Any, Dict
+ExcludeTogglableSecretStr = Annotated[Optional[TogglableSecretStr], Field(exclude=True)]
+"""Annotated type alias for excluding a togglable secret from serialization."""
```
daniel-sintef, line 10:
```
# Defines a type alias that excludes a togglable secret from being serialized.
ExcludeTogglableSecretStr = Annotated[Optional[TogglableSecretStr], Field(exclude=True)]

```

oteapi/strategies/parse/text_csv.py
daniel-sintef: Yeah a bit hard to say, might need a separate issue?

tests/models/test_functionconfig.py
```diff
@@ -31,8 +31,15 @@ def test_functionconfig() -> None:
         "functionType": "foo/bar",
     }
 
+    # NOTE: model_dump_json() returns a JSON string with no whitespace.
```
daniel-sintef, line 34:
```
# NOTE: The method 'model_dump_json()' returns a compact JSON string (no extra spaces or newlines).
#       To match its format, we use 'json.dumps()' with specific parameters to eliminate whitespace.

```

tests/models/test_genericconfig.py
```diff
@@ -85,10 +83,24 @@ def test_attribute_set(generic_config: "CustomConfig") -> None:
     dict key/value-pair), then it should be fine to change the value type for that
     given key.
```
daniel-sintef, line 85:
```
# CAUTION: Starting from Pydantic v2, automatic type coercion for all types is not supported. For instance, 
# directly assigning any type to `str` won't be auto-coerced.
# For such cases, use 'Annotated' to define custom types.
# Example: Directly setting generic_config.configuration["string"] to 3.14 won't convert it to "3.14".
# Refer to the Pydantic documentation on custom types for further details:
# https://docs.pydantic.dev/latest/usage/types/custom/
# UPDATE: In v2.4, a new configuration was introduced to restore this behavior:
# https://docs.pydantic.dev/latest/api/config/#pydantic.config.ConfigDict.coerce_numbers_to_str

```

tests/models/test_genericconfig.py
```diff
@@ -101,42 +113,81 @@ def test_attribute_contains(generic_config: "CustomConfig") -> None:
 
 def test_attribute_del_item(generic_config: "CustomConfig") -> None:
     """Test configuration.__delitem__."""
+    # "float" is not defined as a pydantic model field, i.e., it's not part of the
```
daniel-sintef, line 116:
```
    # The attribute "float" is not part of the standard pydantic model schema;
    # instead, it's an extra attribute in the model.
    # Therefore, when deleted:
    # - It should no longer exist in the model's extras.
    # - It should also be absent from the model instance itself.
```

tests/models/test_genericconfig.py
```diff
@@ -101,42 +113,81 @@ def test_attribute_contains(generic_config: "CustomConfig") -> None:
...
     del generic_config.configuration["float"]
+
     assert "float" not in generic_config.configuration
+    assert "float" not in generic_config.configuration.model_fields_set
+    assert "float" not in generic_config.configuration.model_json_schema()["properties"]
 
-    # "string" is defined as a pydantic model field.
-    # This means it should not be deleted, but reset to its default value
-    # and removed from the set of set fields.
+    # "string" is defined as a pydantic model field, i.e., it's part of the model
```
daniel-sintef, line 129:
```
# The "string" attribute is a standard field within the pydantic model schema.
# Consequently:
# - It should remain part of the model schema and not be deleted from there.
# - However, its value should be removed from the model instance itself.

```

tests/models/test_genericconfig.py
```diff
@@ -101,42 +113,81 @@ def test_attribute_contains(generic_config: "CustomConfig") -> None:
...
 
     del generic_config.configuration["string"]
+
+    assert "string" not in generic_config.configuration
+    assert "string" in generic_config.configuration.model_json_schema()["properties"]
+
+
+def test_attribute_reset_field(generic_config: "CustomConfig") -> None:
+    """Test configuration.reset_field()."""
+    # "float" is not defined as a pydantic model field, i.e., it's not part of the
```
daniel-sintef, line 150:
```
    # The "float" attribute isn't a standard field in the pydantic model schema but is
    # an extra attribute. When reset:
    # - It should be removed from the model's extras.
    # - It should also be absent from the model instance itself.
```

daniel-sintef (approved):
Minor optional suggestions made

oteapi/strategies/parse/text_csv.py
CasperWA: Opened #359 to follow up on this.

oteapi/models/triplestoreconfig.py
CasperWA: The way it's done now, means the type alias will have it's own doc-string in the documentation:
image

tests/models/test_functionconfig.py
CasperWA: The change I've done here has also been replicated for `test_secretconfig.py` and `test_triplestoreconfig.py`.

PR #434, opened 2024-02-27
Added IRIs as extra field properties in datamodels

quaat:
# Description:

## Type of change:

- [ ] Bug fix.
- [ ] New feature.
- [ ] Documentation update.

### DIFF
--- a/oteapi/models/datacacheconfig.py
+++ b/oteapi/models/datacacheconfig.py
@@ -5,18 +5,18 @@ from typing import Optional
 
 from pydantic import Field
 
-from oteapi.models.genericconfig import AttrDict
 
-
-class DataCacheConfig(AttrDict):
+class DataCacheConfig:
     """DataCache Configuration.
-
     This class should not be used directly as a configuration object
     for a strategy object, but only as a configuration field inside
     a configuration object.
     """
 
-    cacheDir: Path = Field(Path("oteapi"), description="Cache directory.")
+    cacheDir: Path = Field(
+        Path("oteapi"),
+        description="Cache directory.",
+    )
     accessKey: Optional[str] = Field(
         None,
         description="Key with which the downloaded content can be accessed. "
@@ -36,6 +36,5 @@ class DataCacheConfig(AttrDict):
     tag: Optional[str] = Field(
         None,
         description="Tag assigned to the downloaded content, typically "
-        "identifying a session. Used with the `evict()` method to clean up a "
-        "all cache entries with a given tag.",
+        "identifying a session. Used with the `evict()` method to clean up all cache entries with a given tag.",
     )
--- a/oteapi/models/filterconfig.py
+++ b/oteapi/models/filterconfig.py
@@ -11,13 +11,22 @@ class FilterConfig(GenericConfig):
     """Filter Strategy Data Configuration."""
 
     filterType: str = Field(
-        ..., description="Type of registered filter strategy. E.g., `filter/sql`."
+        ...,
+        description="Type of registered filter strategy. E.g., `filter/sql`.",
+        IRI="http://purl.org/dc/terms/type",  # type: ignore
     )
-    query: Optional[str] = Field(None, description="Define a query operation.")
+    query: Optional[str] = Field(
+        None,
+        description="Define a query operation.",
+        IRI="http://www.w3.org/1999/02/22-rdf-syntax-ns#Statement",  # type: ignore
+    )
     condition: Optional[str] = Field(
         None,
         description="Logical statement indicating when a filter should be applied.",
+        IRI="http://www.w3.org/2000/01/rdf-schema#comment",  # type: ignore
     )
     limit: Optional[int] = Field(
-        None, description="Number of items remaining after a filter expression."
+        None,
+        description="Number of items remaining after a filter expression.",
+        IRI="http://schema.org/Integer",  # type: ignore
     )
--- a/oteapi/models/mappingconfig.py
+++ b/oteapi/models/mappingconfig.py
@@ -15,6 +15,7 @@ class MappingConfig(GenericConfig):
     mappingType: str = Field(
         ...,
         description="Type of registered mapping strategy.",
+        IRI="http://purl.org/dc/terms/type",  # type: ignore
     )
     prefixes: Optional[Dict[str, str]] = Field(
         None,
@@ -22,8 +23,10 @@ class MappingConfig(GenericConfig):
             "Dictionary of shortnames that expands to an IRI given as local "
             "value/IRI-expansion-pairs."
         ),
+        IRI="http://www.w3.org/2004/02/skos/core#notation",  # type: ignore
     )
     triples: Optional[Set[RDFTriple]] = Field(
         None,
         description="Set of RDF triples given as (subject, predicate, object).",
+        IRI="http://www.w3.org/1999/02/22-rdf-syntax-ns#Statement",  # type: ignore
     )
--- a/oteapi/models/parserconfig.py
+++ b/oteapi/models/parserconfig.py
@@ -8,5 +8,13 @@ from oteapi.models.genericconfig import GenericConfig
 class ParserConfig(GenericConfig):
     """Parser Strategy Data Configuration."""
 
-    parserType: str = Field(..., description="Type of registered parser strategy.")
-    entity: AnyHttpUrl = Field(..., description="IRI to the entity or collection.")
+    parserType: str = Field(
+        ...,
+        description="Type of registered parser strategy.",
+        IRI="http://purl.org/dc/terms/type",
+    )  # type: ignore
+    entity: AnyHttpUrl = Field(
+        ...,
+        description="IRI to the metadata (entity) or collection of entities.",
+        IRI="http://schema.org/URL",
+    )  # type: ignore
--- a/oteapi/models/resourceconfig.py
+++ b/oteapi/models/resourceconfig.py
@@ -21,7 +21,9 @@ class ResourceConfig(GenericConfig, SecretConfig):
     """
 
     resourceType: Optional[str] = Field(
-        None, description="Type of registered resource strategy."
+        None,
+        description="Type of registered resource strategy.",
+        IRI="http://purl.org/dc/terms/type",  # type: ignore
     )
 
     downloadUrl: Optional[HostlessAnyUrl] = Field(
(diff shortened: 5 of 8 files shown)


oteapi/models/datacacheconfig.py
```diff
@@ -10,13 +10,15 @@
 
 class DataCacheConfig(AttrDict):
     """DataCache Configuration.
-
     This class should not be used directly as a configuration object
```
CasperWA, line 13: Revert the deletion, essentially.
```suggestion

    This class should not be used directly as a configuration object
```

oteapi/models/filterconfig.py
```diff
@@ -11,13 +11,22 @@ class FilterConfig(GenericConfig):
     """Filter Strategy Data Configuration."""
 
     filterType: str = Field(
-        ..., description="Type of registered filter strategy. E.g., `filter/sql`."
+        ...,
+        description="Type of registered filter strategy. E.g., `filter/sql`.",
+        IRI="http://purl.org/dc/terms/type",  # type: ignore
```
CasperWA, line 16:
```suggestion
        json_schema_extra={"IRI": "http://purl.org/dc/terms/type"},
```

oteapi/models/filterconfig.py
```diff
@@ -11,13 +11,22 @@ class FilterConfig(GenericConfig):
...
+    )
+    query: Optional[str] = Field(
+        None,
+        description="Define a query operation.",
+        IRI="http://www.w3.org/1999/02/22-rdf-syntax-ns#Statement",  # type: ignore
```
CasperWA, line 21:
```suggestion
        json_schema_extra={"IRI": "http://www.w3.org/1999/02/22-rdf-syntax-ns#Statement"},
```

oteapi/models/filterconfig.py
```diff
@@ -11,13 +11,22 @@ class FilterConfig(GenericConfig):
...
     )
-    query: Optional[str] = Field(None, description="Define a query operation.")
     condition: Optional[str] = Field(
         None,
         description="Logical statement indicating when a filter should be applied.",
+        IRI="http://www.w3.org/2000/01/rdf-schema#comment",  # type: ignore
```
CasperWA, line 26:
```suggestion
        json_schema_extra={"IRI": "http://www.w3.org/2000/01/rdf-schema#comment"},
```

oteapi/models/filterconfig.py
```diff
@@ -11,13 +11,22 @@ class FilterConfig(GenericConfig):
...
     )
     limit: Optional[int] = Field(
-        None, description="Number of items remaining after a filter expression."
+        None,
+        description="Number of items remaining after a filter expression.",
+        IRI="http://schema.org/Integer",  # type: ignore
```
CasperWA, line 31:
```suggestion
        json_schema_extra={"IRI": "http://schema.org/Integer"},
```

CasperWA (changes requested):
Implementing the `IRI` key the way you have is deprecated in pydantic v2, see https://github.com/EMMC-ASBL/oteapi-core/actions/runs/8063034440/job/22024046478?pr=434#step:7:224.

I have made the according suggested changes in one of the files, but this should be repeated throughout all the config files, where `IRI` is implemented/added.
Also, I think you can remove the "type: ignore" comment afterwards.

oteapi/models/datacacheconfig.py
```diff
@@ -36,6 +38,5 @@ class DataCacheConfig(AttrDict):
     tag: Optional[str] = Field(
         None,
         description="Tag assigned to the downloaded content, typically "
-        "identifying a session. Used with the `evict()` method to clean up a "
-        "all cache entries with a given tag.",
+        "identifying a session. Used with the `evict()` method to clean up all cache entries with a given tag.",
```
jesper-friis, line 41: Please ahead to the [PEP8](https://peps.python.org/pep-0008/#maximum-line-length) Python conventions. Max line length is 80 characters.

oteapi/models/filterconfig.py
jesper-friis, line 21: There exists no `rdfs:Statement`. However, there exists a `rdf:Statement`, but is not the correct property to use here. 

I haven't looked into the details of the [data cube](https://www.w3.org/TR/vocab-data-cube/) vocabulary, but  [qb:slice](https://www.w3.org/TR/vocab-data-cube/#ref_qb_slice_LC) seems to be a possible property one could refer to. Slicing is about selecting a subpart of a dataset, witch is exactly the purpose of a filter strategy.

oteapi/models/filterconfig.py
jesper-friis, line 1: I suggested properties for the `query`, `condition` and `limit` fields seems arbitrary chosen. Since there exists no properties in the DCAT and related W3C vocabularies that correspond to these fields, I think that it is better that we define our them in the OTE Interface Ontology (OTEIO).

oteapi/models/mappingconfig.py
```diff
@@ -15,15 +15,18 @@ class MappingConfig(GenericConfig):
...
         description="Type of registered mapping strategy.",
+        IRI="http://purl.org/dc/terms/type",  # type: ignore
     )
     prefixes: Optional[Dict[str, str]] = Field(
         None,
         description=(
             "Dictionary of shortnames that expands to an IRI given as local "
             "value/IRI-expansion-pairs."
         ),
+        IRI="http://www.w3.org/2004/02/skos/core#notation",  # type: ignore
```
jesper-friis, line 26: If using `skos:notation` for prefixes, then we need a to define a custom datatype, like `oteio:prefixType`, such that we can serialise a prefix as:

    :mappingFilter skos:notation "rdfs: <http://www.w3.org/2000/01/rdf-schema#>"^^oteio:prefixType .

However, I think it would be simpler to define our own `oteio:prefix`, such that we can express the above as:

    :mappingFilter oteio:prefix "rdfs: <http://www.w3.org/2000/01/rdf-schema#>" .

oteapi/models/parserconfig.py
```diff
@@ -8,5 +8,13 @@
...
     """Parser Strategy Data Configuration."""
 
-    parserType: str = Field(..., description="Type of registered parser strategy.")
-    entity: AnyHttpUrl = Field(..., description="IRI to the entity or collection.")
+    parserType: str = Field(
+        ...,
+        description="Type of registered parser strategy.",
+        IRI="http://purl.org/dc/terms/type",
+    )  # type: ignore
+    entity: AnyHttpUrl = Field(
```
jesper-friis, line 16: Entity is a difficult name. It can mean anything. I am not found of freely mixing different vocabularies, but if we really want to identify this with `schema:url`, then I think the field should be named `url`.

oteapi/models/resourceconfig.py
```diff
@@ -21,7 +21,9 @@ class ResourceConfig(GenericConfig, SecretConfig):
     """
 
     resourceType: Optional[str] = Field(
-        None, description="Type of registered resource strategy."
+        None,
+        description="Type of registered resource strategy.",
+        IRI="http://purl.org/dc/terms/type",  # type: ignore
     )
 
     downloadUrl: Optional[HostlessAnyUrl] = Field(
```
jesper-friis, line 29: Please spell the field name as `downloadURL` to be consistent with DCAT.

oteapi/models/resourceconfig.py
```diff
@@ -42,6 +45,7 @@ class ResourceConfig(GenericConfig, SecretConfig):
             " type of the distribution is defined in IANA "
             "[[IANA-MEDIA-TYPES](https://www.w3.org/TR/vocab-dcat-2/#bib-iana-media-types)]."
         ),
+        IRI="http://www.w3.org/ns/dcat#mediaType",  # type: ignore
     )
     accessUrl: Optional[HostlessAnyUrl] = Field(
```
jesper-friis, line 50: Please spell the field name as `accessURL` to be consistent with DCAT.

oteapi/models/transformationconfig.py
```diff
@@ -39,40 +39,58 @@ class TransformationConfig(GenericConfig, SecretConfig):
         description=(
             "Type of registered transformation strategy. E.g., `celery/remote`."
         ),
+        IRI="http://purl.org/dc/terms/type",  # type: ignore
     )
     name: Optional[str] = Field(
-        None, description="Human-readable name of the transformation strategy."
+        None,
+        description="Human-readable name of the transformation strategy.",
+        IRI="http://purl.org/dc/elements/1.1/title",  # type: ignore
```
jesper-friis, line 47: Better to use the widely adopted `dcterms` namespace, rather than the old and more limited `http://purl.org/dc/elements/1.1/` namespace.
```suggestion
        IRI="http://purl.org/dc/terms/title",  # type: ignore
```

oteapi/models/transformationconfig.py
```diff
@@ -39,40 +39,58 @@ class TransformationConfig(GenericConfig, SecretConfig):
...
     created: Optional[datetime] = Field(
         None,
         description="Time of creation for the transformation process. Given in UTC.",
+        IRI="http://purl.org/dc/terms/created",  # type: ignore
     )
     startTime: Optional[datetime] = Field(
-        None, description="Time when the transformation process started. Given in UTC."
+        None,
+        description="Time when the transformation process started. Given in UTC.",
+        IRI="http://purl.org/dc/terms/date",  # type: ignore
```
jesper-friis, line 90: `startTime` and `finishTime` cannot have the same IRI. However, DCAT has `dcat:startDate` and `dcat:endDate` which suitable here. 

The only issue with `dcat:startDate` and `dcat:endDate` are that they have domain `dcterms:PeriodOfTime`. Since a `TransformationStatus` describes a time period, but isn't a time periode itself, the RDF serialisation cannot be straight forward, like

```rdf
:transformation_status1 
  dcat:startTime "2024-02-29 08:45" ;
  dcat:endTime "2024-02-29 09:00" .
```

but has to has to be expressed as:
```rdf
:transformation_status1 
  dcterms:temporal [
    a dcterms:PeriodOfTime ;
    dcat:startTime "2024-02-29 08:45" ;
    dcat:endTime "2024-02-29 09:00" ;
  ] .
```

oteapi/models/filterconfig.py
quaat: You are absolutely right about the IRIs being a bit random for certain properties, but I would much rather go to more generic concepts in established ontologies, than making up our own concepts in a custom ontology that nobody has adopted.

